PR #28, opened 2017-03-07
OAuth2 implementation for CalDAV syncing

watersalesman:
I used the original CalDAV script to make another one for syncing with OAuth2 servers, such as Google's CalDAV server. 

In the new script, I changed a few functions such as `remote_query` and `remote_wipe` as well as the authorization functions to work with the new libraries (`httpslib2` and `oauth2client`). Installation of these libraries may be required. I also appended OAuth2 options onto the sample config file.

I tested all use cases (remove local, remove remote, push, pull, etc.) , each type of syncdb init (keep-remote, keep-local, and two-way), and both debug and verbose messages with Google CalDAV. But if anything looks off or I need to document something better, I'd be glad to go back and fix it!

### DIFF
--- a/contrib/caldav/README.md
+++ b/contrib/caldav/README.md
@@ -22,7 +22,7 @@ found under contrib/caldav/config.sample in the calcurse source tree. You will
 also need to install *httplib2* for Python 3 using *pip* (e.g. `pip3 install
 --user httplib2`) or your distribution's package manager.
 
-If you run calcurse-caldav for the first time, you need to provide the --init
+If you run calcurse-caldav for the first time, you need to provide the `--init`
 argument. You can choose between the following initialization modes:
 
     --init=keep-remote Remove all local calcurse items and import remote objects
@@ -51,7 +51,7 @@ How It Works
 ------------
 
 calcurse-caldav creates a so-called synchronization database at
-~/.calcurse/caldav/sync.db that always keeps a snapshot of the last time the
+`~/.calcurse/caldav/sync.db` that always keeps a snapshot of the last time the
 script was executed. When running the script, it compares the objects on the
 server and the local objects with that snapshot to identify items that were
 added or deleted. It then
--- a/contrib/caldav/README.md
+++ b/contrib/caldav/README.md
@@ -65,3 +65,93 @@ added or deleted. It then
 Note: Since calcurse does not use unique identifiers for items, it cannot keep
 track of moved/edited items. Thus, modifying an item is equivalent to deleting
 the old item and creating a new one.
+
+OAuth2 Authentication
+---------------------
+
+calcurse-caldav also has support for services requiring OAuth2 authentication
+such as Google Calendar. Note that you can only have a single calendar synced
+at any given time, regardless of authentication method. To enable OAuth2 you
+will need to:
+
+* Change *AuthMethod* from "*basic*" to "*oauth2*" in your config file
+* Fill in the appropriate settings in your config file: "*ClientID*",
+ "*ClientSecret*", "*Scope*", and "*RedirectURI*". These can be obtained from
+ the API provider. (See below for Google Calendar)
+* Install *oauth2client* for Python 3 using *pip* (e.g. `pip3 install --user
+oauth2client`) or your distribution's package manager
+
+Synchronization With Google Calendar
+-------------------------------------
+
+You will need to use your Google account to create a Google API project and
+enable both the CalDAV API and the Google Calendar API. We will be doing this to
+receive a Client ID and Client Secret. The hostname, path, scope and redirect
+URI are listed below.
+
+First, you will need to go to the [Google Developers Console](https://console.developers.google.com/project) and click *Create
+Project*. After doing that, you can go to the [API Library](https://console.developers.google.com/project/_/apiui/apis/library) and
+search for the CalDAV API and enable it for your project. You will then need to
+do the same for the Google Calendar API.
+
+Next, go to the [Credentials page](https://console.developers.google.com/project/_/apiui/credential)
+, click on *Create credentials*, and choose *OAuth client ID*. If it asks you
+to "set a product name on the consent screen", click on *Configure consent
+screen* to do so. Any product name will do. Upon saving and returning to the
+"Create client ID" screen, choose *Other* as the Application type and click
+*Create*. You now have your Client ID and Client Secret to put into your
+calcurse-caldav config file!
+
+The following options should also be changed in your config file:
+
+```
+Hostname = apidata.googleusercontent.com
+Path = /caldav/v2/*your_calendar_id_here*/events/
+Scope = https://www.googleapis.com/auth/calendar
+```
+
+Your Calendar ID for "*Path*" should be your email for the default calendar.
+If you have multiple calendars, you can get the Calendar ID by going under
+Calendar Details at [calendar.google.com](https://calendar.google.com).
+The default Redirect URI in the config file is http://127.0.0.1; this should
+work fine, but can be changed to http://localhost, a local web server, or
+another device if you encounter errors related to it.
+
+A complete config file for example@gmail.com would have the following options:
+
+```
+[General]
+
+...
+
+Hostname = apidata.googleusercontent.com
+Path = /caldav/v2/example@gmail.com/events/
+AuthMethod = oauth2
+
+...
+
+[OAuth2]
+ClientID = 22802342jlkjlksjdlfkjq1htpvbcn.apps.googleusercontent.com
+ClientSecret = XPYGqHFsfF343GwJeOGiUi
+Scope = https://www.googleapis.com/auth/calendar
+RedirectURI = http://127.0.0.1
+```
+
+[The full guide from Google can be found here.](https://developers.google.com/google-apps/calendar/caldav/v2/guide)
+
+Upon your first run with the `--init` flag, you will be asked to go to a URL to
+log in and authorize synchronization with your Google account. You can access
+this URL on any other device if you cannot open a browser locally (e.g., on
+a headless server). Once you authorize synchronization, you will be redirected
+to your Redirect URI with a code attached to the end, e.g.,
+`http://127.0.0.1/?code=4/Ok6mBNW2nppfIwyL-Q1ZPVkEk3zZdZN3mHcY#`. You will need
+to copy the code after `http://12.0.0.1?code=`. In this case, it would be
+`4/Ok6mBNW2nppfIwyL-Q1ZPVkEk3zZdZN3mHcY#`.
+
+Finally pass this authorization code to calcurse-caldav with the `--authcode`
+flag and initialize the synchronization database like so (note that the quotes
+around the authorization code might be necessary or not, depending on your shell):
+
+```
+calcurse-caldav --init keep-remote --authcode '4/Ok6mBNW2nppfIwyL-Q1ZPVkEk3zZdZN3mHcY#'
+```
--- a/contrib/caldav/calcurse-caldav.py
+++ b/contrib/caldav/calcurse-caldav.py
@@ -10,6 +10,7 @@ import ssl
 import subprocess
 import sys
 import textwrap
+import urllib.parse
 import xml.etree.ElementTree as etree
 
 
@@ -242,7 +243,7 @@ def push_object(conn, objhash):
             etag = next(iter(etagdict.values()))
     etag = etag.strip('"')
 
-    return (href, etag)
+    return (urllib.parse.quote(href), etag)
 
 
 def push_objects(objhashes, conn, syncdb, etagdict):
--- a/contrib/caldav/calcurse-caldav.py
+++ b/contrib/caldav/calcurse-caldav.py
@@ -12,6 +12,14 @@ import textwrap
 import urllib.parse
 import xml.etree.ElementTree as etree
 
+# Optional libraries for OAuth2 authentication
+try:
+    from oauth2client.client import OAuth2WebServerFlow, HttpAccessTokenRefreshError
+    from oauth2client.file import Storage
+    import webbrowser
+except ModuleNotFoundError:
+    pass
+
 
 def msgfmt(msg, prefix=''):
     lines = []
@@ -90,6 +98,65 @@ def get_auth_headers():
     user_password = base64.b64encode(user_password).decode('ascii')
     headers = {'Authorization': 'Basic {}'.format(user_password)}
     return headers
+
+
+def init_auth(client_id, client_secret, scope, redirect_uri, authcode):
+    # Create OAuth2 session
+    oauth2_client = OAuth2WebServerFlow(client_id=client_id,
+                                        client_secret=client_secret,
+                                        scope=scope,
+                                        redirect_uri=redirect_uri)
+
+    # If auth code is missing, tell user run script with auth code
+    if not authcode:
+        # Generate and open URL for user to authorize
+        auth_uri = oauth2_client.step1_get_authorize_url()
+        webbrowser.open(auth_uri)
+
+        prompt = ('\nIf a browser window did not open, go to the URL '
+                  'below and log in to authorize syncing. '
+                  'Once authorized, pass the string after "code=" from '
+                  'the URL in your browser\'s address bar to '
+                  'calcurse-caldav.py using the "--authcode" flag. '
+                  "Example: calcurse-caldav --authcode "
+                  "'your_auth_code_here'\n\n{}\n".format(auth_uri))
+        print(prompt)
+        die("Access token is missing or refresh token is expired.")
+
+    # Create and return Credential object from auth code
+    credentials = oauth2_client.step2_exchange(authcode)
+
+    # Setup storage file and store credentials
+    storage = Storage(oauth_file)
+    credentials.set_store(storage)
+    storage.put(credentials)
+
+    return credentials
+
+
+def run_auth(authcode):
+    # Check if credentials file exists
+    if os.path.isfile(oauth_file):
+
+        # Retrieve token from file
+        storage = Storage(oauth_file)
+        credentials = storage.get()
+
+        # Set file to store it in for future functions
+        credentials.set_store(storage)
+
+        # Refresh the access token if it is expired
+        if credentials.invalid:
+            try:
+                credentials.refresh(httplib2.Http())
+            except HttpAccessTokenRefreshError:
+                # Initialize OAuth2 again if refresh token becomes invalid
+                credentials = init_auth(client_id, client_secret, scope, redirect_uri, authcode)
+    else:
+        # Initialize OAuth2 credentials
+        credentials = init_auth(client_id, client_secret, scope, redirect_uri, authcode)
+
+    return credentials
 
 
 def remote_query(conn, cmd, path, additional_headers, body):
@@ -384,6 +451,7 @@ configfn = os.path.expanduser("~/.calcurse/caldav/config")
 lockfn = os.path.expanduser("~/.calcurse/caldav/lock")
 syncdbfn = os.path.expanduser("~/.calcurse/caldav/sync.db")
 hookdir = os.path.expanduser("~/.calcurse/caldav/hooks/")
+oauth_file = os.path.expanduser("~/.calcurse/caldav/oauth2_cred")
 
 # Parse command line arguments.
 parser = argparse.ArgumentParser('calcurse-caldav')
@@ -402,6 +470,9 @@ parser.add_argument('--syncdb', action='store', dest='syncdbfn',
 parser.add_argument('--hookdir', action='store', dest='hookdir',
                     default=hookdir,
                     help='path to the calcurse-caldav hooks directory')
+parser.add_argument('--authcode', action='store', dest='authcode',
+                    default=None,
+                    help='auth code for OAuth2 authentication')
 parser.add_argument('-v', '--verbose', action='store_true', dest='verbose',
                     default=False,
                     help='print status messages to stdout')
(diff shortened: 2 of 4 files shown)


hughwilliams94:
Would love to give this a try but I am not sure how to begin, I have copied the file in and filled out the config but don't know how to get `calcurse-caldav` to use OAuth2

watersalesman:
@hughwilliams94 I added a new script `calcurse-oauth2.py` in the pull request. Once you've added the options to your config file, run the new script just as you would have with 'calcurse-caldav.py' -- same syntax and all. 

I made it a standalone script because of the amount of changes that had to be made to switch from Basic Authentication to OAuth2, and to be compatible with Google's CalDAV API.

ralpheeee88:
@watersalesman o/ ....followed your instructions as mentioned above. I get the following error

http://ix.io/rv3

(i doubled checked i have <httplib2> installed on my system

watersalesman:
@ralpheeee88 Can you try running the following and seeing if any errors occur?
```
$ python2 -c "import httplib2"
$ python3 -c "import httplib2"
```
The script uses python 3, so it may be that you have `httplib2` installed for python 2 and not python 3.

ralpheeee88:
@watersalesman just trying to find my RedirectURI now

watersalesman:
@ralpheeee88 The default "http://localhost" should work just fine.

The script will open a window in your browser. If it doesn't, it will also provides a link to go to. Login and authorize the application. 

You will then be redirected to something like "localhost/?code=kajsldjf;aj2387sfuoiasj$jlajksdf". Paste everything after "localhost?code=" into your terminal window.

ralpheeee88:
@watersalesman   am just getting 

400. That’s an error.

Error: redirect_uri_mismatch

then it points me back to my api page

watersalesman:
@ralpheeee88 Ah I apologize for that. Change RedirectURI in the config to "http://127.0.0.1" instead of "http://localhost" and let me know if that works.

Referencing localhost performs an actual lookup, so if there is not a record for it in the "hosts" file it will fail. I will fix that.

ralpheeee88:
@watersalesman no worries....i made the changes as mentioned above and still get the same error...(am on my way home so in the next 45 i will be back on line)

ralpheeee88:
@watersalesman chief just wanted to touch base with you and see if there had been any progress

watersalesman:
@ralpheeee88 http://127.0.0.1 still gives you a redirect uri mismatch error? Also, just to confirm, are you trying to sync up with Google Calendar? If so, you may have specified a redirect uri when you registered your API key, so I would ensure that matches up with what you specified there.

ralpheeee88:
@watersalesman  yeah still getting a miss match....confirm google calendar

[https://i.imgur.com/74CxSzd.png](url)

watersalesman:
Interesting. And not to sound like a broken record, (just trying to be absolutely sure) changing that in the configuration file AND on the API side to "http://localhost" does not work either? 

Also a picture of the error when trying both 127.0.0.1 and localhost would be appreciated.

ralpheeee88:
@watersalesman 

Local Host:
https://i.imgur.com/SoBGs22.png
https://i.imgur.com/TtAr933.png

http://127.0.0.1
well i got one step further now....lol
https://i.imgur.com/nFIq3g7.png

ralpheeee88:
@watersalesman https://i.imgur.com/Be6oPAm.png

watersalesman:
@ralpheeee88 Ok, so that went through. The 403 seems like an issue on the API side. Double-check that your calendar has permission to be shared and that you enabled both the CalDAV api and the Google calendar API.

ralpheeee88:
@watersalesman man thxs for your patience....baby steps ....

https://i.imgur.com/tHAFsah.png

watersalesman:
@ralpheeee88 Likewise. I'm learning where I need to document better. 

Can you try running it again to ensure that it wasn't a temporary error on Google's end? That error is generally when a server unavailable due to maintenance or is overwhelmed by traffic at the moment.

ralpheeee88:
@watersalesman  will keep trying ...i assume in the interim i cant use calcurse-caldav

https://i.imgur.com/RYDGJmt.png

watersalesman:
@ralpheeee88 calcurse-caldav and calcurse-oauth2 are independent. Just fill in the config options in the same file as you did previously. Just keep in mind that calcurse can only be synchronized to one calendar at a time, oauth2 or standard caldav.

ralpheeee88:
@watersalesman ok think i am almost there...just waiting for sync to complete....on a side note because i have been trying to sync so many times i have so many appointments duplicated....i just delete the apt file.....and i am getting the following error when closing the calendar

https://i.imgur.com/RSIOcIf.png

once again thxs for all your help!!!

watersalesman:
@ralpheeee88 Ok, I'm glad you got it there. Any feedback on any features needing more documenting, I'd be glad to hear it. 

As for that error, I actually have the same error on both my machines. One of them doesn't have any of the oauth2 code on it. It happened very recently (past couple of days) with updates. It might be worth filing a bug report and watching for new updates.

lfos:
Thanks a lot for working on this and sorry for not replying earlier. This is a nice PoC but not ready for inclusion as-is. When implemented properly, this should be part of the `calcurse-caldav` script itself and the authentication method should be made configurable from the configuration file. Do you want to work on this? :)

watersalesman:
I didn't want to step on your toes by directly integrating, but with your go-ahead, absolutely! I'll work on that.

lfos:
Great! Before you start implementing, a few design decisions to keep in mind:

1. OAuth2 support should be completely optional. The script should still be 100% usuable, even if oauth2client is not even installed (should be as easy as wrapping the `import` statements in a `try` block).

2. We should avoid code duplication as much as possible. If the script needs to be made more flexible to support different authentication methods without copy-pasting code, this should be done in preparatory patches, before OAuth2 support is added.

3. Any interactive input/output should be avoided, if possible. If some initialization is needed, maybe add some more initialization command line flags to keep things flexible.

4. Everything should be documented well and the configuration file prototype should be extended. It would also be nice to have some full tutorial on how to make things work with Google to avoid problems as in GitHub issue #41.

Thanks again for contributing, it is much appreciated!

watersalesman:
I pushed the commits including the integration. I added an *AuthMethod* setting in the config file as well as the appropriate config options under [OAuth2]. OAuth2 is completely optional. All libraries used exclusively by OAuth2 are wrapped in a `try:` block and will not be imported unless *oauth2client* is successfully imported first.

I avoided code duplication by wrapping authentication specific code in checks on the variable *authmethod*. Not too much had to be changed -- only a few things for *httplib2* to work in place of *http*. I also added a couple of functions to run the OAuth2 authorization work flow. 

If an access token is not initialized or the refresh token has become invalid, it will prompt the user with instructions on getting their authorization code and passing it to calcurse-caldav with the `--authcode` flag. This removes the interactive part of the last script. It still opens up the default browser to the authorization screen. This can be removed if you feel that it should not be there. 

Lastly, I updated the README with a tutorial on how to get started with OAuth2 on calcurse-caldav. I also included a tutorial on syncing with Google Calendar, as well as links to the proper resources. I also changed README to README.md just so that Github and other readers would render the Markdown. 

I am open to any feedback or anything that needs to be fixed, even if it is a small styling improvement. I tested the script with OAuth2 using all three init modes, deleting remotely and locally, and adding events remotely and locally. Unfortunately, I do not have a CalDAV server with which I can test basic CalDAV synchronization, but I did not alter any original code. So, it should work exactly as it did previously.

lfos:
Thanks for working on this. I know this is quite a lot of work. This also means that providing a single huge patch is something that is very hard to review. I would prefer having small atomic patches, doing one thing at a time. Also, have short and concise commit messages in imperative mood for each of the commits. See [1] for further details on our guidelines for patch submission (Section 3 probably is the most relevant one, the other ones are mainly describing how to work with Git and how to submit to the ML -- a GitHub PR is just fine).

What I suggest is splitting the work into at least three separate patches:

1. A commit that adds the code changes and update the configuration file.
2. A commit that moves `README` to `README.md` (this is totally unrelated to OAuth2 support).
3. A commit that updates the documentation.

With regard to the implementation, everything looks quite nice already, modulo some nits. I wonder whether the full authentication needs to be done within this script, though? Especially the web browser thing seems awkward (image somebody running this from a headless machine, such as a server, which is not too uncommon). Is this the way that OAuth2 is usually done? Can't we do part of the authentication outside `calcurse-caldav` and just assume that `oauth2_cred` is already there?

[1] http://calcurse.org/files/submitting-patches.html

watersalesman:
I will break up the commit into those three smaller commits. 

`webbrowser.open()` will fail gracefully and simply return False if it cannot open. Although I see your point about it being awkward. Shall I remove it?

As for the rest, this is usually the way OAuth2 works. As far as I am aware, there is no way to avoid logging in using Google's login screen for the first time authenticating -- even if we have the email and password in the config file such as with HTTP basic authentication. OAuth2 is designed to keep the user's login credentials away from the application at all times, only giving an access token that can be revoked at any time. Secure, but definitely a pain for a terminal-based application. A user running in a headless environment would still be able to access the authorization URL on any other device. As long as they get the resulting authorization code back onto their headless device once and pass it to calcurse-caldav with `--authcode`, the refresh token would last indefinitely.

contrib/caldav/README.md
```diff
@@ -49,7 +49,7 @@ How It Works
 ------------
 
 calcurse-caldav creates a so-called synchronization database at
-~/.calcurse/caldav/sync.db that always keeps a snapshot of the last time the
+`~/.calcurse/caldav/sync.db` that always keeps a snapshot of the last time the
```
lfos: All the changes above do not seem to be related to this commit. They deserve a separate patch or could be squashed into the `README.md` patch.

contrib/caldav/README.md
```diff
@@ -63,3 +63,104 @@ added or deleted. It then
 Note: Since calcurse does not use unique identifiers for items, it cannot keep
 track of moved/edited items. Thus, modifying an item is equivalent to deleting
 the old item and creating a new one.
+
+OAuth2 Authentication
+---------------------
```
lfos: Style nit: Missing empty line after the heading.

contrib/caldav/README.md
```diff
@@ -63,3 +63,104 @@ added or deleted. It then
...
+
+Example for distro where Python 3 is the default (e.g., Arch Linux):
+```
+pip install --user oauth2client
+```
+Example for distro where Python3 is not the default (e.g., Fedora 26):
+
+```
+pip3 install --user oauth2client
+```
```
lfos: Can't we always advise to use `pip3` here? Then maybe just put this into the bullet point above, like "Install *oauth2client* for Python 3 using *pip* (e.g. `pip3 install --user oauth2client`) or your distribution's package manager."

contrib/caldav/README.md
```diff
@@ -63,3 +63,104 @@ added or deleted. It then
...
+
+Synchronization With Google Calendar
+-------------------------------------
```
lfos: Missing newline.

contrib/caldav/README.md
```diff
@@ -63,3 +63,104 @@ added or deleted. It then
...
+screen* to do so. Any product name will do. Upon saving and returning to the
+"Create client ID" screen, choose *Other* as the Application type and click
+*Create*. You now have your Client ID and Client Secret to put into your
+calcurse-caldav config file!
+
+The following options should also be changed in your config file:
+
+* Hostname = apidata.googleusercontent.com
+* Path = /caldav/v2/*your_calendar_id_here*/events/
+* Scope = https://www.googleapis.com/auth/calendar
```
lfos: Do we want to `quote` these?

contrib/caldav/README.md
```diff
@@ -63,3 +63,104 @@ added or deleted. It then
...
+```
+calcurse-caldav --init keep-remote --authcode 4/Ok6mBNW2nppfIwyL-Q1ZPVkEk3zZdZN3mHcY#
+```
+
+If you are using a shell other than bash, such as zsh, you may need to surround
+the authorization code with single quotes:
+
+```
+calcurse-caldav --init keep-remote --authcode '4/Ok6mBNW2nppfIwyL-Q1ZPVkEk3zZdZN3mHcY#'
+```
```
lfos: Again, can't we just always advise to use the version with quotes? Maybe add an extra comment, like "Finally pass this authorization code to calcurse-caldav with the `--authcode` flag and initialize the synchronization database like so (note that the quotes around the authorization code might be necessary or not, depending on your shell):"

lfos:
My question about the authentication was rather like: Can the initial authentication steps (until we have that `oauth2_cred` file) be done by another command-line tool? I am by no means an OAuth2 expert, so if this is the standard way of doing things, I am fine with it.

I started reviewing the documentation and I like it so far. I added some comments above. Thank you again for your contribution and patience!

watersalesman:
From my limited experience with OAuth2, integrating initial authentication into the program itself is definitely the standard -- especially since it is already handling the opening and storage of the credentials file (`oauth2_cred`) and refreshing of tokens. 
 
Great, I will make those changes. I'm glad I can contribute!

contrib/caldav/calcurse-caldav.py
```diff
@@ -92,6 +102,65 @@ def get_auth_headers():
...
+        webbrowser.open(auth_uri)
+
+        prompt = ('\nIf a browser window did not open, go to the URL '
+                  'below and log in to authorize syncing. '
+                  'Once authorized, pass the string after "code=" from '
+                  'the URL in your browser\'s address bar to '
+                  'calcurse-caldav.py using the "--authcode" flag. '
+                  "Example: calcurse-caldav --authcode 'your_auth_code_here'\n\n{}\n"
+                  .format(auth_uri))
+        print(prompt)
```
lfos: Maybe wrap this single long line at a sensible width?

contrib/caldav/calcurse-caldav.py
```diff
@@ -132,7 +211,6 @@ def remote_query(conn, cmd, path, additional_headers, body):
         die(("The server at {} replied with HTTP status code {} ({}) " +
              "while trying to access {}.").format(hostname, resp.status,
                                                   resp.reason, path))
-
```
lfos: Unrelated whitespace change.

contrib/caldav/calcurse-caldav.py
```diff
@@ -182,7 +260,12 @@ def remote_wipe(conn):
     if dry_run:
         return
 
-    remote_query(conn, "DELETE", path, {}, None)
+    if authmethod == 'basic':
+        remote_query(conn, "DELETE", path, {}, None)
+    elif authmethod == 'oauth2':
+        remote_items = get_etags(conn)
+        for href in remote_items:
+            remove_remote_object(conn, remote_items[href], href)
```
lfos: This surprised me quite a bit. You do we need special handling of OAuth2 here?

contrib/caldav/calcurse-caldav.py
```diff
@@ -112,15 +181,25 @@ def remote_query(conn, cmd, path, additional_headers, body):
...
+        body = resp.read().decode('utf-8')
 
-    headers = resp.getheaders()
-    body = resp.read().decode('utf-8')
+    elif authmethod == 'oauth2':
+        # httplib2 can separate headers and body through simultaneous assignment
+        # Using headers and resp variables for compatibility
+        headers, body = conn.request(path, cmd, body=body, headers=headers)
+        resp = headers
+        body = body.decode('utf-8')
```
lfos: Out of curiousity: Can't we just use httplib2 for both basic and OAuth2 authentication. Then we could drop some of this magic. If so, it might be a good idea to switch basic authentication to httplib2 in a preparatory ptach first, then add OAuth2 support.

contrib/caldav/calcurse-caldav.py
```diff
@@ -233,7 +316,10 @@ def push_object(conn, objhash):
         return None
 
     etag = None
-    headerdict = dict((key.lower(), value) for key, value in headers)
+    if authmethod == 'basic':
+        headerdict = dict((key.lower(), value) for key, value in headers)
+    elif authmethod == 'oauth2':
+        headerdict = dict((key.lower(), headers[key]) for key in headers)
```
lfos: Same as above. Looks like `headers` has a different type here, depending on whether OAuth2 is used or not. This might be superfluous if we switch basic authentication to use httplib2 as well first. But even if we do not, we should make `headers` always be a proper dictionary when it is set instead. Maybe a simple cast is enough.

contrib/caldav/calcurse-caldav.py
```diff
@@ -242,6 +328,10 @@ def push_object(conn, objhash):
             etag = next(iter(etagdict.values()))
     etag = etag.strip('"')
 
+    if authmethod == 'oauth2':
+        # Return href to a format recognized by the syncdb by removing hostname and encoding
+        href = urllib.parse.quote(uri_end + objhash + ".ics")
+
```
lfos: Again, could you please explain why this is needed? I would not have expected any special code in the synchronization routines...

contrib/caldav/calcurse-caldav.py
```diff
@@ -263,7 +353,10 @@ def push_objects(objhashes, conn, syncdb, etagdict):
 
 def remove_remote_object(conn, etag, href):
     headers = {'If-Match': '"' + etag + '"'}
-    remote_query(conn, "DELETE", href, headers, None)
+    if authmethod == 'basic':
+        remote_query(conn, "DELETE", href, headers, None)
+    elif authmethod == 'oauth2':
+        remote_query(conn, "DELETE", http_host + href, headers, None)
```
lfos: Same as above.

contrib/caldav/config.sample
```diff
@@ -33,3 +36,16 @@ Verbose = Yes
...
+#[OAuth2]
+#ClientID = your_client_id
+#ClientSecret = your_client_secret
+
+# Scope of access for API calls. Synchronization requires read/write
+#Scope = https://example.com/resource/scope
+
+# Change the redirect URI if you receive errors, but ensure that it is identical
+# to the redirect URI you specified in the API settings
+#RedirectURI = http://127.0.0.1
```
lfos: Just a minor note: The other comments always end with a period. Maybe add that to the new comments as well for consistency. Thanks!

contrib/caldav/calcurse-caldav.py
watersalesman: I am not sure if this is an issue specific to the httplib2/Google Calendar, but when pushing objects, the URI path is saved to the syncdb without URL encoding. The next time syncing, it will delete everything pushed locally and pull it from the server, because it sees the encoded version from the server and the unencoded version locally as two different objects. I prevented this from happening by adding in this bit. At the very least I should change the comment. It is not very helpful in understanding why that line is there.

contrib/caldav/calcurse-caldav.py
watersalesman: While the original method of remote wiping is significantly faster, it is unfortunately not supported by Google Calendar. Rather than sending a single DELETE request to the calendar collection root, it requires that each event be deleted individually. It is probably something related to security/best practices.

I can add a comment here to document that.

contrib/caldav/calcurse-caldav.py
watersalesman: You're right. I had a bit of a brain fart and could have just done a simple cast, as you mentioned, to convert it from an httplib2.Response object to a proper dict. I will fix this.

contrib/caldav/calcurse-caldav.py
lfos: Sounds like something which is not related to OAuth2 then. So we should probably:

1. Switch to making separate requests for both basic authentication and OAuth2.
2. Do it in a separate patch.

contrib/caldav/calcurse-caldav.py
lfos: Also sounds like something that should be investigated and fixed properly (in a separate patch, since it is completely unrelated to OAuth2) :)

contrib/caldav/calcurse-caldav.py
watersalesman: Ok, I will do that.

contrib/caldav/calcurse-caldav.py
watersalesman: It definitely appears to have all the needed functionality that *http.client* has, including disabling SSL certificate validation for basic authentication. While it would definitely help clean things up, it comes with the downside of adding a dependency to basic authentication usage. It comes down to whether the refactor is worth adding that dependency.

lfos:
Generally speaking, checking that the OAuth2 patch only touches the authentication part of the script a good sanity check to ensure that commits are split properly... Everything addressing other issues should be explained and done separately.

contrib/caldav/calcurse-caldav.py
watersalesman: Because of the way *httplib2* takes URI parameters (absolute URI as opposed to a URI path), the href need to be passed along with the entire beginning of the URL e.g. "https://host.name.com" + "/href/". I could rename the variables to be a little more clear. Or this could be eliminated if we replace *http.client* with *httplib2*

watersalesman:
I will split up those pre-OAuth2 patches into separate commits. If we switch to httplib2 entirely, we would also have to update the documentation to state that as a dependency.

lfos:
If it really makes the code simpler and more consistent, I'd say go for it. That additional dependency probably does not hurt too much...

watersalesman:
Sorry for a bit of a delay...still getting used to some of the slightly more advanced features of git. 

I included those pre-OAuth2 patches, including the refactor for httplib2, into separate commits before the addition of OAuth2 support. I also included the other miscellaneous fixes.

lfos:
Great! Everything looks good to me now. I merged your patches to `pu` and will let them cook for a while. Will close this PR and get back to you if there are any issues.

Thanks a lot for this nice piece of work!

PR #47, opened 2017-09-25
make delete key act as delete instead of backspace

### LINKED ISSUE #34: Delete acts as backspace
I noticed it when I was putting in an appointment. I used the delete key to delete right of the cursor, but it deleted to the left (as backspace does).

I am using calcurse 4.2.2 on Arch Linux.

cryptolmath:
This fixes #34.


lfos:

Issue #48, opened 2017-09-28
[Feature Request] Calcurse-caldav external password support


refacto:
I really like the functionality of calcurse-caldav, the only problem I have is that I don't like storing my password in a plain text file.
What I'm requesting for is that it should be possible to get the password for calcurse-caldav via external command, so that it'd be possible to issue something like 'pass ls calendar'. That way, passwords aren't stored in an unencrypted fashion.

lfos:
You can now set the password using the `CALCURSE_CALDAV_PASSWORD` environment variable; see 9be2c10 (calcurse-caldav: Read password from env variable, 2017-11-10) for details.

PR #49, opened 2017-09-30
Add Command Line Argument For Basic Authentication Password

watersalesman:
This addresses issue #48 

The optional "--password" argument overrides the corresponding option in the config file if the argument is present. This allows for passwords to be passed securely -- for example:

```
calcurse-caldav --password $(pass ls calcurse)
````

If anything seems off, or if you don't agree with the feature/design, just let me know. Thanks!


refacto:
Would be nice to do this from withing the config, but this already helps 👍

lfos:

dasJ:
Downside: The password is visible in top and ps. I think it would be more secure if you could pass it via an env variable like `CALCURSE_PASSWORD=$(pass show calcurse) calcurse-caldav`

watersalesman:
@dasJ Good point. I'll put up a pull request soon to add that and document it.

PR #50, opened 2017-10-06
Add another readline shortcut: CTRL('G')

hiqua:


lfos:

Issue #55, opened 2017-10-24
CalDAV syncing working except for tasks


ramblingenzyme:
I've setup Google Calendar syncing with calcurse-caldav, and it's working great except that if I have TODO tasks, it tries and fails to sync them with a 403 on apidata.googleusercontent.com/caldav/user@gmail.com/events/something.ics

watersalesman:
Hmmm interesting. What kind of task is this and how did you go about creating it? I only ask because I have synced successfully despite having reminders and tasks (it ignores them since they are not calendar events). If you let me know how you made it, I can go about reproducing and troubleshooting the error.

ramblingenzyme:
Sorry, I don't think I made it clear enough that I'm talking about todo items in calcurse made locally. If I have any todo items in calcurse, the sync script tries to upload them to Google Calendar and fails on an endpoint it gets a 403 on.

watersalesman:
Ok I understand now. I reproduced it on my end.

The problem is that, while TODO items are exported in an appropriate iCalendar/CalDAV format, Google just isn't having it on their end since Google Calendar does not support TODO items at all.

I will look into how the script could avoid exporting TODO items when syncing using OAuth2.

ramblingenzyme:
https://github.com/lfos/calcurse/blob/efd76a0d995292e48f5466fccada4901618f7d97/contrib/caldav/calcurse-caldav.py#L77 is where the issue comes from. For the moment, I've just removed it from my local version.

I think the correct way would be to add a config option under [General] to disable syncing todos, and allow users to disable it as needed.

lfos:
PR #56 has been merged; you can use the `SyncFilter` option to only synchronize events/appointments now.

paindespik:
It would be great if it could sync the todo with google tasks

PR #56, opened 2017-10-25
Add config option to disable syncing of TODOs

ramblingenzyme:
To address https://github.com/lfos/calcurse/issues/55 particularly for Google Calendar syncing.

### DIFF
--- a/contrib/caldav/calcurse-caldav.py
+++ b/contrib/caldav/calcurse-caldav.py
@@ -44,6 +44,11 @@ def die_atnode(msg, node):
     die(msg)
 
 
+def validate_sync_filter():
+    valid_sync_filter_values = {'event', 'apt', 'recur-event', 'recur-apt', 'todo', 'recur', 'cal'}
+    return set(sync_filter.split(',')) - valid_sync_filter_values
+
+
 def calcurse_wipe():
     if verbose:
         print('Removing all local calcurse objects...')
@@ -53,13 +58,20 @@ def calcurse_wipe():
 
 
 def calcurse_import(icaldata):
-    p = subprocess.Popen([calcurse, '-i', '-', '--dump-imported', '-q',
-                          '--format-apt=%(hash)\\n',
-                          '--format-recur-apt=%(hash)\\n',
-                          '--format-event=%(hash)\\n',
-                          '--format-recur-event=%(hash)\\n',
-                          '--format-todo=%(hash)\\n'],
-                         stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+    command = [
+        calcurse,
+        '-i', '-',
+        '--dump-imported',
+        '-q',
+        '--filter-type', sync_filter,
+        '--format-apt=%(hash)\\n',
+        '--format-recur-apt=%(hash)\\n',
+        '--format-event=%(hash)\\n',
+        '--format-recur-event=%(hash)\\n',
+        '--format-todo=%(hash)\\n'
+    ]
+
+    p = subprocess.Popen(command, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
     return p.communicate(icaldata.encode('utf-8'))[0].decode('utf-8').rstrip()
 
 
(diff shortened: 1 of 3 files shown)


watersalesman:
This does not address the issue in #55 as is. It actually breaks TODO syncing altogether by attempting to export the objects without formatting them.

Take a look at some of the filtering flags in the man page to get the right functionality. Also the new option should probably be added to the example config file and maybe the documentation.

ramblingenzyme:
Yup, I realised that...I've now used the filter flags and added documentation.

lfos:
How about, instead of adding a `SyncTodo` option, allowing for setting the `filter-type` value from the configuration file?

lfos:
Maybe even allow for specifying the whole import command? Not saying that it is a good idea, just something I wanted to throw in for discussion...

lfos:
I plan on releasing 4.3.0 in the next couple of days, by the way. If we want this feature in there, we should try to find a concensus soon...

watersalesman:
I am in favor of the filter-type config option. Makes the script more flexible going forward (e.g. if someone wants to use basic auth but does not want to sync TODOs and recurring events, or this case where TODOs cause an error with Google Calendar).

ramblingenzyme:
I'm happy to get on this by tomorrow when I'm a bit more free.

contrib/caldav/calcurse-caldav.py
```diff
@@ -43,6 +43,12 @@ def die_atnode(msg, node):
         msg += etree.tostring(node).decode('utf-8')
     die(msg)
 
+def validate_sync_filter():
+    valid_sync_filter_values = {'event', 'apt', 'recur-event', 'recur-apt', 'todo', 'recur', 'cal'}
+
+    filter_items = set(sync_filter.split(','))
+
+    return filter_items - valid_sync_filter_values
```
lfos: Small nit: I think this is easier to read if you make it a two-liner, initializing `valid_sync_filter_values` in the first line and computing the return value (i.e. `return set(sync_filter.split(',')) - valid_sync_filter_values`) in the second line, without an empty line inbetween.

lfos: We also separate functions by two empty lines. So there is one empty line missing before the function and another one after the function.

contrib/caldav/calcurse-caldav.py
```diff
@@ -53,13 +59,20 @@ def calcurse_wipe():
...
-                          '--format-apt=%(hash)\\n',
-                          '--format-recur-apt=%(hash)\\n',
-                          '--format-event=%(hash)\\n',
-                          '--format-recur-event=%(hash)\\n',
-                          '--format-todo=%(hash)\\n'],
-                         stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+    command = [
+        calcurse,
+        '-i', '-',
+        '--filter-type', sync_filter,
```
lfos: Minor nit: This should go after, `--dump-imported` and `-q`, just before the formatting options.

contrib/caldav/config.sample
```diff
@@ -18,6 +18,13 @@ Path = /path/to/calendar/on/the/server/
 # Enable this if you want to skip SSL certificate checks.
 InsecureSSL = No
 
+# This option allows you to filter the type of tasks synced and the valid options are
+# event,apt, recur-event, recur-apt, todo, recur, cal exactly like the command line option
```
lfos: Nit: Missing space after `event,`. Maybe also mention that there should not be any whitespace in the comma-separated list itself.

lfos:
I decided to not include this in 4.3.0, so there is no need to rush. I plan on releasing 4.4.0 soon after 4.3.0, though, because there is a lot of WIP on `calcurse-caldav`. The plan is to make proper support for both Google Calendar and iCloud the milestone to reach for 4.4.0.

lfos:
Merged with some small changes to the commit message and to the comment in the sample configuration file. Thanks!

Issue #30, opened 2017-04-13
Connection with iCloud


coreice:
First, thanks for making available this script.

Did someone already tried making a CalDAV connection to the iCloud? I've been trying but without succes so far.

`warning: Dry run; nothing is imported/exported. Add "DryRun = No" to the
warning: [General] section in the configuration file to enable synchronization.
Connecting to caldav.icloud.com...
Removing all local calcurse objects...
error: The server at caldav.icloud.com replied with HTTP status code 400 (Bad
error: Request) while trying to access //.`

The config file:
`# If you want to synchronize calcurse with a CalDAV server using
# calcurse-caldav, create a new directory ~/.calcurse/caldav/, copy this file
# to ~/.calcurse/caldav/config and adjust the configuration below.

[General]
# Path to the calcurse binary that is used for importing/exporting items.
Binary = calcurse

# Host name of the server that hosts CalDAV.
Hostname = caldav.icloud.com

# Path to the CalDAV calendar on the host specified above.
Path =       

# Enable this if you want to skip SSL certificate checks.
InsecureSSL = No 

# Disable this option to actually enable synchronization. If it is enabled,
# nothing is actually written to the server or to the local data files. If you
# combine DryRun = Yes with Verbose = Yes, you get a log of what would have
# happened with this option disabled.
DryRun = Yes

# Enable this if you want detailed logs written to stdout.
Verbose = Yes

# Credentials for HTTP Basic Authentication. Leave this commented out if you do
# not want to use authentication.
[Auth]
Username = mark@heralder.net
Password = edoe-cwdk-nbln-oluq

# Optionally specify additional HTTP headers here.
#[CustomHeaders]
#User-Agent = Mac_OS_X/10.9.2 (13C64) CalendarAgent/176`

lfos:
What version of the script are you using? Did you try current master? Does running the script with `--debug` show anything helpful?

xkpd3:
Hi,
I'm also playing with icloud right now. The issue you have is that you did not provide a path.
Finding the path is tricky, as you need the user ID.

The full path would be something like this:
```
https://caldav.icloud.com/USER_ID/calendars/CALENDAR_ID
```

This tool might help you with that: https://github.com/muhlba91/icloud

Currently, I'm as far as that calcurse-caldav syncs with this URL:
```
https://caldav.icloud.com/USER_ID/calendars/
```
[...]

Issue #46, opened 2017-09-21
All day events are exported as zero duration events


hiqua:
I export my calcurse calendar to the ical format using `calcurse -x`. I then import it into the Android calendar. The problem is that what are called "all-day event"s in calcurse are just events occurring at 00:00 in Android, so I don't see them anymore after 00:01.

When looking at the ical file exported by calcurse, it appears that there is no DTEND field for these events. This is the correct behaviour for events that do not take time according to RFC 2445 ([cf Recommended pratices, 2](https://tools.ietf.org/html/rfc2445#page-140)).
For events that actually take the whole day, DTEND should be specified as DTSTART+1 day (or +n days depending on how many days it lasts), since DTEND specificies the non-inclusive ending time.

calcurse already supports events with only a starting date (just fill the starting date and leave the ending date empty). I think it would be better if an all-day event really had a duration of 24h.

Overall I'm not sure whether calcurse's behaviour is wrong per se, but always writing a DTEND for all day events would remove the ambiguity.

Related: it would be nice to be able to create all-day events spanning several days; for now, it's only possible to specify a duration if a precise time has been chosen.

lfos:
See PR #60 for a related issue -- iCloud does not seem to like events without `DTEND` at all. So we should try to improve the way `DTEND` is exported for all-day events and for punctual appointments (maybe add a configurable workaround specifically for iCloud if we agree that not specifying `DTEND` for punctual appointments is the right thing to do)...

hiqua:
I just realized that even though the lack of DTEND field makes calcurse differ from what Android expects, the real reason why all day events are not considered as such in Android is the difference in the DTSTART field. calcurse exports these fields as:
`DTSTART:20171001`
while it actually works when I change this field to this format (which is the one used when exporting from Android):
`DTSTART;VALUE=DATE:20171001`

So this is actually a bug in the app I use to import the calcurse calendars (ICSDroid), even though it would work correctly, were calcurse to export using the explicit format with the VALUE field.

lfos:
Should be fixed by now. Please reopen if the issue persists.

Issue #20, opened 2017-01-23
calcurse-caldav: synchronization problem


nwtti:
I'am able to initialize the Database, but if i try to sync it, calcurse-caldav can't read the initialized database file. My Server is running Baikal 0.5.0

```
Connecting to ///SERVER///...
Loading synchronization database from /home/witti/.calcurse/caldav/sync.db...
Traceback (most recent call last):
  File "/usr/bin/calcurse-caldav", line 519, in <module>
    syncdb = get_syncdb(syncdbfn)
  File "/usr/bin/calcurse-caldav", line 196, in get_syncdb
    href, etag, objhash = line.rstrip().split(' ')
ValueError: not enough values to unpack (expected 3, got 1)
```

lfos:
This definitely sounds like a bug to me. Can you reproduce this?

shapesinspace:
Just installed calcurse 4.2.2 initialized database for calcurse-caldav and am experiencing the same error ....Loading synchronization database from ~/user/.calcurse/caldav/sync.db...
Traceback (most recent call last):
  File "/usr/local/bin/calcurse-caldav", line 519, in <module>
    syncdb = get_syncdb(syncdbfn)
  File "/usr/local/bin/calcurse-caldav", line 196, in get_syncdb
    href, etag, objhash = line.rstrip().split(' ')
ValueError: not enough values to unpack (expected 3, got 2)

lfos:
Sounds like your synchronization database is defective. There is probably some line in `~/.calcurse/caldav/sync.db` that contains no spaces (every line should consist of three tokens separated by spaces). Could you check and let me know what the offending line looks like? Maybe also have a look at the line before and after.

shapesinspace:
sync.db contained;
'/dav/calendars/user/shapesinspace@fastmail.net/788c1e89-2580-4556-8337-         
b47ef7a89f64/db9vo55es0jlkvi3adohoulm4c@google.com.ics                          
ab813d7a9ac84cd7c90733c6e0bf620942cedaee
Some items could not be imported, see log file? [y/n]'

i removed the line and grepped for the associated entry. found the entry and then deleted it from the remote server and corresponding local entry. reran calcurse-caldav --init=keep-remote followed by calcurse-caldav. Same error resulted as I had originally posted. This time I found offending entries in sync.db, different from the previous time;
2 double entries and two of the same 'some items could not be imported..' messages;

[...]

PR #60, opened 2017-11-01
Adding iCloud Support

xkpd3:
This is my first attempt to make `calcurse-webdav` play with the Apple iCloud calendar. It's supposed to solve issue https://github.com/lfos/calcurse/issues/30.

...but not yet. The download works, sync works, upload does not yet work.

I'll post this pull request already. It would be interesting to know if the download still works with other caldav servers. Does it break google?

Issues:
- iCloud sends the base URL as resource: **Solved** by ignoring resources that do not end in "/"
- iCloud does not allow REPORT without an VEVENT filter: **Solved** by refactoring the download to use PROPFIND instead of REPORT.
- iCloud sends multiple events in one .ical resource. Unfortunately the .ical resource is the key in the database: **Solved** by concatenating the UID spearated by ";" and saving them on one .ical resource. Expanding the list to figure out what needs to be synced.
- iCloud expects a special UID form in a PUT request: To be done
- iCloud can not handle DTSTART without DTEND: To be done


watersalesman:
This is some good work! I'm looking forward to seeing the increased compatibility. 

I'm not sure if it is specific to syncing with Google Calendar or not, but OAuth2 syncing fails with the new changes. I receive the following error: https://gist.github.com/watersalesman/bb6bb6e3557977407904c9937ad172aa

This happens with normal syncing, as well as any of the --init types.

xkpd3:
@watersalesman Thanks for testing. Looks like I need to revive my google account and test this with iCloud side by side. Interesting, that I read apple, calendar server, mobileme etc in your error trace. This is not coming from `calcurse-webdav`... I doubt it's coming from google. Is your google calendar somehow connected to icloud or did you migrate from icloud to google in the past? I'm curious...

watersalesman:
@xkpd3 Nope, this account is in no way shape or form connected to any Apple products/accounts. A [quick search](https://stackoverflow.com/questions/37655843/google-caldav-bad-request-on-put-event) shows other people getting things such as mobileme within their responses from Google Calendar, so it's likely from there.

lfos:
Thanks so much for working on this. I know this is a non-trivial task and requires quite some patience and debugging skills, so your work is much appreciated!

Some remarks:

* While putting all into a single commit is fine for a draft/RFC, we should really split each of the points addressed into a single commit with a much more detailed explanation in the commit message when merging this. We can either do the splitting later or you can do it while you are working on it -- whatever you prefer.
* I would like to avoid switching from `REPORT` to `PROPFIND` unless it is really necessary (mainly because of potential breakages with other servers that were supported before). Is there really no way to tune the `REPORT` query such that iCloud CalDAV server likes it? What is the exact error message you get when filtering for a combination of `VEVENT` and `VTODO`, alongside with `test="anyof"`?
* The fix for "multiple events per file" is a very good one. It definitely needs to go in a separate commit (even more than the others) because I think we also saw this issue with other servers but could not pinpoint the cause -- thanks for looking into this and coming up with a fix. With regard to the implementation, I personally prefer `,` to `;` as a separator but that is probably just a matter of taste.
* The "special UID" thing is something to investigate in more details before "fixing". Again, I would really try to avoid any changes here unless absolutely necessary because of potential regressions.
* As mentioned in the other bug report, we need to discuss what to do with `DTEND` when exporting all day events and punctual appointments. Issue #46 is related.

lfos:
You are totally right about `anyof`. It is only in "Collected Extensions to CalDAV". Not sure how I missed that. Let us do two separate calls then. And yes, the `DTEND` issue needs to be fixed in the export routine of calcurse itself.

xkpd3:
- Split commit into smaller chunks: ✓
- Use multiple `REPORT` calls instead if `PROPFIND`: ✓

I can see a weird sync bahvior right now. Download works fine, When a local event gets added it gets synced. The next sync will remove a lot of local entries even they have not been removed remotely.

I assume that the syncdb expand must be done in multiple places. I tried to move it to the `get_syncdb` function, but this won't work as it breaks `href` as key. So the expansion must be done when accessing the data afte is has been selected via href key.

lfos:
Well, currently, the internal data structure used to represent the synchronization database is a dictionary which maps each URL to an ETag and to a calcurse object hash. To implement the "multiple events per URL" feature properly, I guess we should change this to be a dictionary of 2-tuples, where the first entry in the tuple is the ETag and the second entry is a *list* of calcurse object hashes.

Then, the expansion of lists needs only be done when reading (writing) the synchronization database from (to) the disk. And, of course, one must check all uses of the `syncdb` variable in the script and make adjustments where required.

lfos:
Regarding the multuple `REPORT` calls, I would like to reduce code duplication as much as possible. These two queries are almost the same, so it makes sense to have a single hard-coded template and insert the requested event type using `format()`.

xkpd3:
Reg. your first comment:
Changing the db structure this way, would this allow to acces data by hash as well as by href? If yes, this would be a good move.

Reg. your second comment:
I'll put it on the todo list :)

lfos:
Depends on what you mean by "allow to acces data by hash as well as by href". Of course, the data structure allows accesses in both directions because it contains all information from the synchronization database. Lookups by `href` are more efficient, though, while each access by object hash requires O(n) time (linear scan). I think this is not noticable, unless you have thousands of appointments, though.

dasJ:
So, is it safe to test this?

lfos:
I don't think so. This is pretty incomplete and, as Stefan mentioned, currently broken...

xkpd3:
@dasJ: I'm a bit short with free time right now. I'll continue on the weekends when possible.

The next step is a bigger one, because it involves a database structure change.

Which google calendar bug are you referring to? Maybe we can fix this earlier.

dasJ:
@xkpd3 Thank you for your work! There is already #20 which seems to be my problem

lfos:
Maybe we can do smaller steps. The directory href patch can be included as-is, I guess. I renamed issue #20 -- maybe we can fix that next in a separate PR.

lfos:
I added some ideas and open problems to #20 -- it would be great to have more people join that discussion!

lfos:
What's the status of this PR? Do you plan on continuing your work on this?

xkpd3:
I'm planning to continue, I just have trouble finding the time for it.

lfos:
Any progress? Would you like somebody else to pick this up?

xkpd3:
If anybody wants to pick this up - please do so. I can only reproduce the issues with my real calendar. And I lost many entries because of it. I was not able to reproduce the issue in my test calendar. This multi-event-entries never showed up.

lfos:
The problem is that I cannot reproduce the behavior at all... Maybe somebody else can help isolate the issue?

rslindee:
Is this stalled because the developer is unable to reproduce? Please let me know if this is the case and I can test on my 100% consistently-failing setup.

lfos:
It is stalled because the original PR submitter stopped working on it. Missing pieces are described in the comments above. Feel free to pick it up.

vxid:
Note that [vdirsyncer](https://vdirsyncer.pimutils.org/en/stable/tutorials/icloud.html) supports icloud remotes (see #188)

lfos:
Closing due to inactivity. I would be really happy if somebody with iCloud access picks this up and completes the work though!

PR #61, opened 2017-11-05
BSD friendly shebang

xkpd3:


lfos:
Thanks. I would appreciate a better commit message. This has nothing to do with *BSD per se. The commit message should also indicate that the patch affects `calcurse-caldav`. How about `calcurse-caldav: Use /usr/bin/env for python3` or `calcurse-caldav: Honor PATH environemnt`, alongside with a more detailed explanation in the message body? While you're at it, it would also be nice if you could sign-off the patch by using the `-s` argument when running `git commit`.

xkpd3:
I've change the commit accordingly. Thanks.

lfos:
I changed the `[calcurse-caldav]` prefix to `calcurse-caldav:` because that is what we already used for all the past commits, and I also fixed a typo in the commit message, and merged the result into `pu`. Thanks again!

Issue #59, opened 2017-10-31
Feature request (?): Multiple CalDAV calendars


mrossinek:
Hi!

I successfully managed to synchronize my calcurse installation with my private nextcloud server.
Now I am wondering, if it is possible to sync more than one calendar with the same calcurse profile.. So far, I can only sync the one calendar specified within the Path of my calcurse-caldav config file.

I'd be happy to hear any tipps or a possible implementation of this feature.

Cheers!

lfos:
You can use different configuration files and pass them to `calcurse-caldav` using the `--config` command line option, see issue #14 for more information... Does that solve your issue?

mrossinek:
It solves it partially I guess.
I am able to also specify separate sync database files using the `--syncdb` command line option.
However, whenever initializing the sync of one or the other calendar all items get deleted locally..
A possible solution would be to add another option to initialize each synchronisation. I.e. keep both

However I guess this only becomes feasible once calcurse may support several calendars because as of now this would sync all items into each of the calendars.

lfos:
You can use `--init=keep-local` if you want to do a one-way import from the server. Syncing with two different calendars (different data directories and different sync DBs) should work perfectly fine right now. You only cannot access them at the same time (i.e. using the same running instance) with `calcurse`.

mrossinek:
Okay thanks for clarifying!
I will close this issue then.

PR #64, opened 2017-11-05
[PoC] Support for multiple accounts

xkpd3:
Hi,

This is another "work in progress" pull request. I just want to get some feedback about this.

I have restructured the config file. Some settings are now account dependent.

```
# Configure account sections for multiple calendar support
Accounts = icloud-calendar-stefan, icloud-todos-stefan

[icloud-calendar-stefan]
AuthMethod = basic
Type = calendar
Hostname = p50-caldav.icloud.com
Path = /1365501036/calendars/M2CD-2-2-43DB4CA6-FDB9-4636-B859-BC42A5C2DF5C/
Username = hugo123
Password = *****
# ClientID                                                                                                                                                     
# ClientSecret
# Scope
# RedirectURI

[icloud-todos-stefan]
AuthMethod = basic
Type = todo
Hostname = p50-caldav.icloud.com
Path = /1365501036/calendars/5173B391-DA95-4B0F-9A74-25C4477DFC50
Username = fred123
Password = ******
# ClientID
# ClientSecret
# Scope
# RedirectURI
```

In the `Accounts` setting, you can put a comma separated list of accounts that will be configured later on.
Then add a new section for each of the accounts and configure them. Calcurse-caldav should then do several runs and sync one after the other.

Additionally you can explicitely set a `Type`. This is not yet implemented, but should allow you to tell calcurse-caldav what should be synced with the configure ressource. The type will also help with calendar servers that can not fetch multiple resource types with one request (iCloud for example).

*BUG ALERT!*
Don't try this without dryrun yet!!

@lfos: calcurse-caldav can be aware about what has been synced with which resource. Calcurse core does not. This becomes a problem when entries are created in calcurse. Then we don't know to which calendar we should sync the entry. Would it be possible to add an account identitier that will be saved together with each entry?

Alternatively we can set a default calendar and whatever gets created locally, goes to this calendar.

### DIFF
--- a/contrib/caldav/calcurse-caldav.py
+++ b/contrib/caldav/calcurse-caldav.py
@@ -92,9 +92,9 @@ def calcurse_version():
 
 
 def get_auth_headers():
-    if not username or not password:
+    if not auth_data[account][2] or not auth_data[account][3]:
         return {}
-    user_password = ('{}:{}'.format(username, password)).encode('ascii')
+    user_password = ('{}:{}'.format(auth_data[account][2], auth_data[account][3])).encode('ascii')
     user_password = base64.b64encode(user_password).decode('ascii')
     headers = {'Authorization': 'Basic {}'.format(user_password)}
     return headers
@@ -151,10 +151,10 @@ def run_auth(authcode):
                 credentials.refresh(httplib2.Http())
             except HttpAccessTokenRefreshError:
                 # Initialize OAuth2 again if refresh token becomes invalid
-                credentials = init_auth(client_id, client_secret, scope, redirect_uri, authcode)
+                credentials = init_auth(oauth_data[account][0], oauth_data[account][1], oauth_data[account][2],oauth_data[account][3], authcode)
     else:
         # Initialize OAuth2 credentials
-        credentials = init_auth(client_id, client_secret, scope, redirect_uri, authcode)
+        credentials = init_auth(oauth_data[account][0], oauth_data[account][1], oauth_data[account][2],oauth_data[account][3], authcode)
 
     return credentials
 
@@ -169,7 +169,7 @@ def remote_query(conn, cmd, path, additional_headers, body):
     headers.update(additional_headers)
 
     if debug:
-        print("> {} {}".format(cmd, path))
+        print("> {} {}".format(cmd, account_data[account][1]))
         print("> Headers: " + repr(headers))
         if body:
             for line in body.splitlines():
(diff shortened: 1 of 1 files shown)


lfos:
I am not sure whether I like the idea of adding "synchronization types", see pull request #56 for a different approach. Also, what is the benefit of adding multiple account support (vs. using different configuration files)? What is the expected behavior when specifying multiple accounts with the same synchronization type? Will they all be synchronized and merged, i.e. they all share exactly the same set of items after synchronization?

xkpd3:
Hi @lfos,

It's independent functionality. I thought you might want to have this in separate requests. The main idea behind this is, that you can sync multiple calendar resources with calcurse.

It would enable you to sync with several google calendars as well as maybe an icloud calendar at the same time.

I have a work calendar, a personal calendar and a shared calendar and it would be cool to have them all in calcurse merged together. This PR is supposed to make this work.

I have seen https://github.com/lfos/calcurse/pull/56 and it would be obsolete with this PR. However, I consider this PR to be much more powerful because it would enable you to sync calendar entries with google and todos with apple. Or maybe even pull together all todos from apple and google.

For now, it's more a proof of concpet and I think we should have exactly this conversation. Is it something you would like calcurse to be able to do? Or are you happy with the limit of supporting just one calendar.

lfos:
I still fail to see what you think this should work. Suppose you have a personal calendar and a work calendar. You are right that we can make `calcurse-caldav` aware of which imported items originate from which CalDAV calendar to avoid copying the appointments of each of the calendars into all other calendars. However, how should `calcurse-caldav` know which of the CalDAV calendars to import new appointments (created within calcurse itself) into?

What you can already do is use different configuration files for `calcurse-caldav` and separate calcurse data files to mirror your personal calendar and your work calendar locally. However, you will not be able to display items from both sources at once in calcurse.

For proper support, we probably need support for multiple calendars within calcurse itself (as suggested in issue #59) first.

xkpd3:
More interesting in this context is https://github.com/lfos/calcurse/issues/59, which describes how to sync different servers with different config files.

I did not see this before and I think this may "good enough" and keeps the code much cleaner.

xkpd3:
Calcurse caldav would know from which calendar which entry was synced. The differrent accounts would have different paths and this can be used to avoid that everything gets synced everywhere. But yes, three is the problem that when calcurse hands back an objhash that's not known to calcurse-caldav. This woud then go to the default calendar.

I did not yet solve all problems. To fully support this, we would need to hand a calendar ID to calcurse that is stored alongside the entry and handed back to calcurse-caldav on update. The next demand then would be to allow the user to choose the target calendar within calcurse.

As I said. I guess I went down this path without considerung the alternative route to alias calcurse with difffernt config files and work this way. I think this works just as well.

xkpd3:
I'll close this PR. I think https://github.com/lfos/calcurse/issues/59 and https://github.com/lfos/calcurse/pull/56 in combination are flexible enough.

PR #66, opened 2017-11-07
contrib/caldav/README.md: Fix localhost IP address and URI formatting

rjc:


lfos:
Good catch. The commit message should be something like `calcurse-caldav: Fix localhost IP address` and it is missing your sign-off (which can be created by passing `-s` to `git commit`).

lfos:
Do you plan on resubmitting with fixed commit message and a `Signed-off-by` line? :)

rjc:
I've sent a patch to the bugs@ mailing list instead :^)

PR #67, opened 2017-11-08
Add options to read password from environment

watersalesman:
An extension of #49. As @dasJ mentioned, it is more secure to pass it through an environment variable as it wouldn't show up in something like top or ps. 

I also documented the external password support.

### DIFF
--- a/contrib/caldav/calcurse-caldav.py
+++ b/contrib/caldav/calcurse-caldav.py
@@ -508,6 +508,9 @@ authcode = args.authcode
 verbose = args.verbose
 debug = args.debug
 
+# Read environment variables
+password = os.getenv('CALCURSE_CALDAV_PASSWORD')
+
 # Read configuration.
 config = configparser.RawConfigParser()
 if verbose:
@@ -563,10 +566,8 @@ if config.has_option('Auth', 'UserName'):
 else:
     username = None
 
-if config.has_option('Auth', 'Password'):
+if config.has_option('Auth', 'Password') and not password:
     password = config.get('Auth', 'Password')
-else:
-    password = None
 
 if config.has_section('CustomHeaders'):
     custom_headers = dict(config.items('CustomHeaders'))
--- a/contrib/caldav/README.md
+++ b/contrib/caldav/README.md
@@ -32,6 +32,14 @@ argument. You can choose between the following initialization modes:
 For subsequent calcurse-caldav invocations, you don't need to specify any
 additional parameters.
 
+You can specify a username and password for basic authentication in the
+config file. Alternatively, the password can be passed securely from another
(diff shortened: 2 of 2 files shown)


lfos:
I am not convinced it is a good idea to support both the environment variable and the command line option. Maybe we should drop the patch adding the command line option in favor of the environment variable?

watersalesman:
I agree with dropping the previous patch. No reason to keep both.

lfos:
Could you rebase the patch on master (i.e. avoid that the config file password overwrites the environment variable), please?

watersalesman:
Done. I dropped efd76a0d995292e48f5466fccada4901618f7d97 and added the tidbit that allows the external password to override the config option.

contrib/caldav/calcurse-caldav.py
```diff
@@ -508,6 +508,12 @@ def run_hook(name):
 verbose = args.verbose
 debug = args.debug
 
+# Read environment variables
+try:
+    password = os.environ['CALCURSE_PASSWORD']
+except KeyError:
+    password = None
+
```
lfos: I think it is much cleaner (and shorter) to use `password = os.getenv('CALCURSE_PASSWORD')` here.

lfos:
Ew, I just realized that the commit is already part of master (confused it with the `SyncFilter` patch before), and even part of a release, so we cannot drop it anymore. I will revert the patch instead and apply the two patches (once you addressed the detail I pointed out above) on top instead.

watersalesman:
Made the edit. Should I undo the rebase or leave it as is?

lfos:
No need to do anything with the rebase, I cherry-picked those two patches. I noticed that the commit message of the second patch talks about a `CALCURSE_CALDAV` environment variable, while the variable itself is called `CALCURSE_PASSWORD`. That actually made me think of whether `CALCURSE_PASSWORD` is a good name because it suggests that it stores some password for calcurse... Maybe `CALCURSE_CALDAV_PASSWORD` or `CALCURSE_CALDAV_PASS` instead? I know that it is quite verbose but I cannot think of anything shorter which describes the purpose well enough and does not seem too artificial.

watersalesman:
Good point. `CALCURSE_CALDAV_PASS` sounds like the best one to me.

lfos:
I think I even like `CALCURSE_CALDAV_PASSWORD` a bit better since it matches the configuration option name. It is quite long but then again, we only save four letters when using `CALCURSE_CALDAV_PASS` instead...

lfos:
By the way, I reverted the commit adding the command line option in master, so you can rebase on master in the next round.

watersalesman:
Edited the variable name and rebased

lfos:
Merged into `pu`. Thanks!

Issue #94, opened 2018-04-17
"Unsupported Media Type" error when exporting all-day events through calcurse-caldav


maxigaz:
When I create all-day events in calcurse and try to sync them with a Nextcloud calendar using calcurse-caldav, I get the following error message:

```
Connecting to cloud.disroot.org...
Loading synchronization database from /home/xxx/.calcurse/caldav/sync.db...
Pushing new object d2c91457a872abffe7935a222537fc430051f190 to the server.
error: The server at cloud.disroot.org replied with HTTP status code 415
error: (Unsupported Media Type) while trying to access https://cloud.disroot.org
error: /remote.php/dav/calendars/<username>/<calendar-id>
error: /d2c91457a872abffe7935a222537fc430051f190.ics.
```

If I delete those events newly created in calcurse, replace them with events with specific start and end times, and run calcurse-caldav again, it completes without errors.

Another thing I noticed is that there's no issue when all-day events created in the Nextcloud web interface or Android's calendar are imported into calcurse. They just appear with an asterix in front of them in the list:

```
* Example event created outside calcurse
  Example event created in calcurse
```

I don't know if it's related to #46, so I'm creating a separate issue for now.

Using calcurse v4.3.0 under Manjaro Linux.

lfos:
It looks like the first two dates originate from all-day events and the others from appointments (with a starting time). Isn't that the case?

hiqua:
@loweryaustin can you try by using the format I mentioned in #81, by adding `;VALUE=DATE`:

```
       DTSTART;VALUE=DATE:20070628
       DTEND;VALUE=DATE:20070709
```

Just to be sure that solving #81 would actually solve your issue with nextcloud (that I do not use personally).

loweryaustin:
Sure thing, I have to wait till I get home to try it. Probably should hear back from me around 16:00 CST.

loweryaustin:
@hiqua I tried exporting the event with: `calcurse -x ical --export-uid --filter-hash=28013ef7148261da9f566d617edfd648f7920592 > test.ical`
I then tried importing it through the NextCloud GUI interface. No matter what I did to try to import it I ran into a different validation error that is unrelated to this issue, so I don't really have a good way to test your change.

hiqua:
Sorry my request was confusing. What I meant is: can you try to add ';VALUE=DATE' manually in your `test.ical` file, by editing the file?

To obtain:
```
 DTSTART;VALUE=DATE:20180501
[...]

Issue #81, opened 2017-12-20
ical export is not RFC 5545 compliant (DTSTART field)


hiqua:
ICSDroid does not parse the ical export from calcurse correctly, so I reported it on the ICSDroid side, but Ricki Hirner from ICSDroid pointed me to the following which indicates that the bug is actually in calcurse.

According to RFC 5545 (https://tools.ietf.org/html/rfc5545#section-3.8.2.4)
```
3.8.2.4.  Date-Time Start

   Property Name:  DTSTART

   Purpose:  This property specifies when the calendar component begins.

   Value Type:  The default value type is DATE-TIME.  The time value
      MUST be one of the forms defined for the DATE-TIME value type.
      The value type can be set to a DATE value type.
```
and (https://tools.ietf.org/html/rfc5545#section-3.3)
```
3.3.  Property Value Data Types

   The properties in an iCalendar object are strongly typed.  The
   definition of each property restricts the value to be one of the
   value data types, or simply value types, defined in this section.
   The value type for a property will either be specified implicitly as
   the default value type or will be explicitly specified with the
   "VALUE" parameter.  If the value type of a property is one of the
   alternate valid types, then it MUST be explicitly specified with the
   "VALUE" parameter.
```
which means that whenever a DATE field (instead of DATE-TIME) has to be specified for DTSART (or DTEND), it has to be specified with the VALUE tag, such as:
```
       DTSTART;VALUE=DATE:20070628
       DTEND;VALUE=DATE:20070709
```
instead of:
```
       DTSTART:20070628
       DTEND:20070709
```
as is currently the case. These examples are from the RFC 5545 as well.

Although calcurse mentions RFC 2445, it is superseded by RFC 5545, so I think this is relevant.

Once the issue is acknowledged, I can create a PR.

hiqua:
I guess this means that you agree about the issue, so yeah I'll try to work on this, it looks easy enough for me.

lfos:
Fixed by 0114289 (ical: use the VALUE parameter in DTSTART, 2018-05-09).

PR #97, opened 2018-05-09
ical: use the VALUE parameter in DTSTART

hiqua:
Solves #81, possibly #94.

I have only tested this on my own calendar, more testing would be appreciated, especially wrt #94.


lfos:
Merged to `pu`, thanks!

PR #103, opened 2018-05-25
Transparent cursor fix

lhca:
The (Fix no-colour theme)-patch partly destroyed the cursor in getstring() by
turning it into a solid block. The fix reintroduces wchgat() which requires a
color pair argument. When no colors are wanted, color pair 0 is used.

A similar problem exists in the layout and colour customization windows and
is fixed in the same way: move to the position and apply the reverse video
attribute.


lfos:
Merged into `pu` with a slightly amended commit message, thanks!

PR #104, opened 2018-05-25
Scrollbar and right window border (corrected)

lhca:
When a scrollbar is on display in APP or TOD windows, the right
vertical border (outside the scrollbar) is not highlighted
when the window is selected.

The scrollbar is always highlighted:
- when APP or TOD is deselected
- in configuration windows where borders otherwise are not

The patch moves the scrollbar parameters (except highlight)
from arguments of draw_scrollbar() to the function itself.
The highlight argument was 1; instead it is set higher in
the call hierarchy (wins_update_panels()) and passed on down.

### DIFF
--- a/src/calcurse.h
+++ b/src/calcurse.h
@@ -170,6 +170,9 @@
 /* Size of the hash table the note garbage collector uses. */
 #define NOTE_GC_HSIZE 1024
 
+/* Mnemonics */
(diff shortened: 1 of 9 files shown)


lfos:
I still don't like the looks of the black scroll bar on a black-on-white color theme but it is probably better than what we have now. Merged into `pu`, thanks!

Issue #86, opened 2018-02-18
config.c: 589: configuration variable unknown: "appearance.headingpos"


adjunkth:
I use calcurse on my phone and on my computer and use Dropbox to sync the files:

Computer - Debian 9
calcurse version 4.2.1

Android 6.0
Termux
Calcurse version 4.3.0

The problem occurs when I have changed the files on my phone and tries to use the same files on my computer. However if I open up my conf file and erase the following two lines (Which seems to have been added by calcurse version 4.3.0):

appearance.headingpos=right-justified
format.dayheading=%B %-d, %Y

Then everything works fine. 

If I don't erase the files I get the following error message:
config.c: 589: configuration variable unknown: "appearance.headingpos"

and can not open the program

hiqua:
In summary, either unknown configuration variables should be ignored instead of calcurse exiting, or the configuration format should not change between minor versions.

lfos:
... or you should not synchronize configuration files accross different releases (sharing data files is okay).

hiqua:
I never thought about using calcurse in termux, but I think I'll end up doing it as well, when it becomes possible.

Is there a way to specify the path of the configuration file without changing the path of the data files? (so `-D` does not work)

I have tried to use `-c` to specify the apts file, but the notes are still in the data directory, so it's not a good solution either.

I think being able to specify some configuration folder (which would also contain keys) different from the data folder would be a suitable workaround, and the changes should be rather minimal. It would then be easier to just sync the whole data folder, and use device-specific configuration files.

This assumes that the data (not configuration) format does not change, at least not too often.

@lfos if you're open to this I can look into it.

adjunkth:
Yes of course

I like calcurse and would like to see it work. Calcurse works well in termux 
and that sounds like a good solution



onsdag 9 maj 2018 kl. 04:13:19 CEST skrev  Quentin Hibon:
> I never thought about using calcurse in termux, but I think I'll end up
> doing it as well, when it becomes possible.
> 
> Is there a way to specify the path of the configuration file without
> changing the path of the data files? (so `-D` does not work)
> 
> I have tried to use `-c` to specify the apts file, but the notes are still
> in the data directory, so it's not a good solution either.
> 
> I think being able to specify some configuration folder (which would also
> contain keys) different from the data folder would be a suitable
> workaround, and the changes should be rather minimal. It would then be
> easier to just sync the whole data folder, and use device-specific
> configuration files.
> 
[...]

PR #105, opened 2018-05-27
Add option to specify the configuration file used

hiqua:
The configuration file (~/.calcurse/conf by default) can now be specified with
-C or --conf.

Workaround for GitHub issue #86.

Let me know if more changes are required.

### DIFF
--- a/src/args.c
+++ b/src/args.c
@@ -83,7 +83,7 @@ enum {
 static void usage(void)
 {
 	printf("%s\n", _("usage: calcurse [--daemon|-F|-G|-g|-i<file>|-Q|--status|-x[<format>]]\n"
-			 "                [-c<file>] [-D<path>] [-h] [-q] [--read-only] [-v]\n"
+			 "                [-c<file>] [-C<file] [-D<path>] [-h] [-q] [--read-only] [-v]\n"
 			 "                [--filter-*] [--format-*]"));
 }
 
@@ -120,6 +120,7 @@ static void help_arg(void)
 	putchar('\n');
 	printf("%s\n", _("Miscellaneous:"));
 	printf("%s\n", _("  -c, --calendar <file>   Specify the calendar data file to use"));
+	printf("%s\n", _("  -C, --conf <file>       Specify the configuration file to use"));
 	printf("%s\n", _("  --daemon                Run notification daemon in the background"));
 	printf("%s\n", _("  -D, --directory <path>  Specify the data directory to use"));
 	printf("%s\n", _("  -g, --gc                Run the garbage collector and exit"));
@@ -404,17 +405,19 @@ int parse_args(int argc, char **argv)
 	int xfmt = IO_EXPORT_ICAL;
 	int dump_imported = 0, export_uid = 0;
 	/* Data file locations */
-	const char *cfile = NULL, *datadir = NULL, *ifile = NULL;
+	const char *cfile = NULL, *datadir = NULL, *ifile = NULL,
+	      *conffile = NULL;
 
 	int non_interactive = 1;
 	int ch;
 	regex_t reg;
 
-	static const char *optstr = "FgGhvnNax::t::d:c:r::s::S:D:i:l:qQ";
+	static const char *optstr = "FgGhvnNax::t::C:d:c:r::s::S:D:i:l:qQ";
 
 	struct option longopts[] = {
 		{"appointment", no_argument, NULL, 'a'},
 		{"calendar", required_argument, NULL, 'c'},
+		{"conf", required_argument, NULL, 'C'},
 		{"day", required_argument, NULL, 'd'},
 		{"directory", required_argument, NULL, 'D'},
 		{"filter", no_argument, NULL, 'F'},
@@ -474,6 +477,9 @@ int parse_args(int argc, char **argv)
 			break;
 		case 'c':
 			cfile = optarg;
+			break;
+		case 'C':
+			conffile = optarg;
 			break;
 		case 'd':
 			if (is_all_digit(optarg)) {
@@ -716,7 +722,7 @@ int parse_args(int argc, char **argv)
 	else if (to >= 0 && range >= 0)
 		EXIT_IF(to >= 0, _("cannot specify a range and an end date"));
 
-	io_init(cfile, datadir);
+	io_init(cfile, datadir, conffile);
 	io_check_dir(path_dir);
 	io_check_dir(path_notes);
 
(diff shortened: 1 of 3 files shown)


hiqua:
I think it's better indeed, I just didn't want to break compatibility with previous versions. If this is a desirable change (and we break compatibility), I can implement it. Let me know.

lfos:
I am not sure what the benefit of putting the configuration file in its own directory would be. If you want to change the notes directory, you can already do so by using `-D` and then override the configuration file path using `-C`.

hiqua:
I understood that @lhca would prefer to have a directory containing not only conf, but also keys and hooks, while another directory would contain apts, todo and notes. The assumption would be that if you want to change the path to conf, it would be consistent to have keys and hooks at the same place.

I'll let @lhca argue further in favor of this change if its advantage can be made clearer.

lhca:
I just reread @hiqua's original suggestion, and mine is in fact the same and will, of course, break compatibility. It could be left for a future release.

What was the original problem? Sharing calcurse data files between installations of calcurse with different releases is OK for calendar data, but not for configuration data. Hence it is desirable to be able to have different configuration data files and to tell which one to load. This has been solved for the conf file by adding the -D option.

But keys and hooks are also configuration data. If you run calcurse on a stationary computer, a laptop and a smartphone (termux) you may well want different key bindings (even when using the same calcurse version).

Furthermore, isn't there a problem with the -c option? Suppose two different installations of calcurse use the same data directory (no -D option involved), but one use the -c option to specify a different appointments file. They still share the notes. Indeed, use of -D could solve this, but the point is that such a setup should not be allowed.

I am trying to anticipate problems by keeping together what logically belongs together.

lhca:
Please elaborate. How is this compatible? Doesn't `-D` have a new meaning?

How do the new defaults `~/.calcurse/data` and `~/.calcurse/config` come into existence?

lfos:
I did not realize that `-D` is already in use. We can just introduce a new long option `--data-dir` instead and keep the current meaning of `-D`.

The new defaults `~/.calcurse/data` and `~/.calcurse/config` are used by default when calcurse creates a new directory layout. On already existing directories, the semantics I described above is used.

lhca:
@lfos wrote:

> The new defaults ~/.calcurse/data and ~/.calcurse/config are used by default when calcurse creates a new directory layout

I.e. they are created if they do not exist (new directory layout). But you also wrote:

> If neither is specified, the data directory defaults to ~/.calcurse/data unless this directory does not exist, in which case it defaults to ~/.calcurse. Similarly, the configuration directory defaults to ~/.calcurse/config unless this directory does not exist, in which case it defaults to ~/.calcurse.

I.e. if they do not exist, the default is the good old `~/.calcurse`.

lfos:
No, maybe I should be more precise. If `~/.calcurse/` does not exist, it is initialized with the "new" layout. Otherwise, the data directory defaults to `~/.calcurse/data/`, unless that directory does not exist in which case it defaults to `~/.calcurse/`. Similarly, the configuration directory defaults to `~/.calcurse/config/`, unless that directory does not exist in which case it defaults to `~/.calcurse/`.

Thinking about it again, maybe we could even switch `-D` to the new data directory semantics. But then, we would have to switch the config directory order to default to the data directory if it is specified, then to `~/.calcurse/config/` and fall back to `~/.calcurse/` if the latter does not exist. This behavior might be a bit too awkward and confusing...

lfos:
Alternatively, we could simply switch `-D` to the new data directory semantics and let it default to `~/.calcurse/data/` (fall back to `~/.calcurse/`). Analogously, `-C` could be used to specify to config directory and it defaults to `~/.calcurse/config/` (fall back to `~/.calcurse/`). Then, instead of doing the weird hack suggested above, we could check whether the data directory contains the `apts` file and tell users to run `calcurse-upgrade` in that case.

The `calcurse-upgrade` script could be extended to move the files under `~/.calcurse/` to comply with the new layout.

This might be a bit cleaner but requires some user intervention. It makes sure all existing setups are migrated to the new layout, though, which might be desirable. What do you think?

lhca:
That you have made my head spin. Back to start. How about this:

By default all calcurse files are kept together as now (either in `~/.calcurse` or in the directory specified by `-D`. The new `-C <dir>` option means "The conf, keys and hooks files are in directory `<dir>`, but  everything else is where it always was".

The only difference from @hiqua's solution is that the `-C` option specifies a directory, not a file, and that _all_ configuration files, not just `conf`, are there. With the `-C` option present, the old data directory becomes our new data directory, and any configuration files present there are ignored.

lfos:
I like the idea; I still wonder whether we should split configuration files and data files into two separate subdirectories (as a long-term goal, not necessarily before the next release).

@hiqua, do you want to work on a patch to change the semantics of `-C` as suggested?

hiqua:
Yes I will, but maybe it makes sense to revert the changes that introduced `-C` in the meantime. Up to you.

I think it should be doable within 2 months, I'll keep you posted.

lhca:
One further twist. The directory argument to `-C` could be made optional, the default being `~/.calcurse/config/`.

In this way "calcurse" and "calcurse -C" switches between the the two default configuration directories with the data directory staying the same.

lhca:
Oops! That should of course have been:

The directory argument to `-C` could be made optional, the default being the `config/` subdirectory of the data directory whether that is the one specified by `-D` or the default one.

lfos:
Not sure. This kind of suggests that having stuff in a separate `config/` subdirectory is something special or at least something we want to support in parallel to keeping everything in the same directory. If we want to split out config files, we should look into making that the default.

hiqua:
So `config/` by default when creating new files, and just use them as usual if they're not in `config/`. `-C` has a required argument rather than an optional one.

lhca:
I'm not so easily convinced. @lfos wrote:

> If we want to split out config files, we should look into making that the default.

Why? You didn't give a reason. It certainly adds complexity. As it is today, the calcurse file paths are initialised from

* the default values for the calcurse directory (`~/.calcurse`), subdirectories and file names
* the invocation parameters

Nothing else is needed. Simple and robust.

If the default values are changed, backwards compatibility comes into play, and the old defaults must be taken into account as noted by @hiqua:

* the existence of new and old default files/directories in the file system

Presumably the `-D` option must also have a new interpretation (config subdirectory) with accompanying backward compatibility.

The original aim was served. Why change default values?

hiqua:
Let's move this discussion to a proper issue (#111) which is more visible than this closed pull request. Maybe other users will chime in with things we haven't thought about so far.

PR #106, opened 2018-05-28
Check for empty string in config_parse_int()

lhca:
We had an email discussion about the semantics of is_all_digit() and what it should return on an empty string.

I ended accepting your view, but the correction in config_parse_int() was never implemented.


lfos:
Great. Merged into `pu`, thanks!

PR #109, opened 2018-05-29
Utf8 final

lhca:
The final touches (hopefully) to the UTF-8 character keybindings which should make them much more usable.

Also support for translations in po-files: user choices may use single- as well as multibyte characters.


lfos:
Pushed with some minor style changes; thanks!

PR #110, opened 2018-06-04
Appointment fixes

lhca:
The changes are minimal on purpose.

There is a more substantial patch series upcoming.

PR #113, opened 2018-06-07
Date time validation

lhca:
Please excuse this somewhat lengthy comment. I could not find a better way to submit it with the pull request. It is the promised follow-up to lfos/calcurse#110.

# On entering date and time
Lars Henriksen <LarsHenriksen@get2net.dk>

The patch series address several issues which arose from two commits (introduced in release 4.2.0):

- 1f39b5c Add support for moving items to another day
- adf29c8 Support dates when specifying start/end times

They made it possible, when creating or changing an appointment or event, in addition to (or instead of) start and end _time_ to enter a _date_. They also made it possible to move an appointment.

These very useful enhancements also raised some issues which are discussed below together with possible solutions.

### Issue

You may enter a date in addition to a time as mentioned above. If none is given the currently selected day is understood.

If you set up an apppointment/event for a day different from the currently selected one, there is no immediately visible effect on the screen. Hence, you cannot be sure that your input was as you intended it. You will have to move to the date you entered to ascertain that you got what you wanted and did not make a mistake (typing or otherwise). Also, you might want to check whether the new entry clashes with existing ones.

If, in fact, you did make a typing mistake in the (otherwise valid) date, you may have difficulty in locating the entry if you are not aware of your mistake.

Some date short forms are accepted. If year or month is omitted, the current year/month is understood. If you intend to enter 0015 (a quarter past midnight), but by mistake enter 015, what you intended as a time is interpreted as a date and you get an event for the 15. of the current month. [Much more serious confusions were possible before commit "e9bddf3 Parse time before date in parse_datetime()".]

In short, whenever you use the ability to insert an item on a day different from the currently selected one, you invariably move to the day afterwards.

### Solution

I will argue that the natural behaviour is that _first_ you move in the calendar to the day of interest, and _then_ you insert an entry for that day, and not the other way around. If that is accepted, no date entry is needed in this particular case.

Hence, when a new appointment or event is inserted, no date is allowed in the start time. In all other cases (including the end time of a new appointment) an optional date may be entered if needed.

### Issues

The prompt texts in the status bar do not reflect the possibility of entering a date, nor do they outline date input formats.

When existing start/end times are changed, only the time, and not the date, is presented for the user to edit in the status bar. This is particularly annoying for the end time which may be on another day.

The status bar texts explaining time and duration formats are becoming rather cluttered.

### Solution

Status bar texts are updated with dates and date formats.

Already existing start and end times are presented in full (date and time) for editing.

When there is insufficient space on the first status bar line, built-in help (in the status bar) is provided when "?" is typed. User input is temporarily interrupted, and both status bar lines are used to display help information. After acknowledgement from the user, input is resumed.

### Issue

__NOTE.__ In the following "date" means a point in time (including hour, minutes and seconds) specified as number of seconds before or after the Unix epoch (midnight January 1, 1970))

Dates entered into calcurse must of course be valid in the application context (start before end), but also in the operating system context.  The operating system puts limits on the allowed time span. The limits depend on the size of the data type (`time_t`) used for dates: 32 or 64 bits.

Calcurse must check any date (entered as text) for consistency with the OS, before accepting it as valid.

Date calculations must guard against aritmethic overflow (wrap around).

### Solution

Some of these issues were addressed by commit "691d6b3 Check for the year span 1902-2037".

All dates are checked for OS validity.

Overflow detection for addition and multiplication is implemented.

When an appointment is moved, the new start is validated, but also the implied new end date.

### Database inconcistency

It is possible to get an inconsistent database with an appointment that ends before it begins:

- edit an existing appoitment
- specify an end time by a date that precedes the start time

Or

- create a new appointment
- give any time as start time
- give a date as end time that lies before the start

You will have to exit/start calcurse to get the error message

`apoint.c: 229: date error in appointment`

[It is no longer enough to save/reload the data because the reload is a silent noop since "57dd3d6 Only reload if data files were changed".]

### Event detection

__Note.__ This is not related to the date issues.

If the start time of a new appointment is rejected as invalid in the first attempt, the new calendar entry becomes an event (no end time).

.Example:
- add an item in the calendar,
- for start time enter "x",
- after the rejection correct "x" to a valid time.
- the end time question is skipped and you are asked to enter a description.

### Solution

Bugs are fixed.


lfos:
Sorry, it took me quite a while to review this whole series and fix the merge conflicts but I am happy to say that this is now part of `pu`! I will let it cook there for a while before merging into master. Thanks for all the great work!

PR #114, opened 2018-06-10
Fix multiple, simultaneous periodic saves

lhca:
A new save thread was started every time a positive periodic save value was input
in the general options configuration menu. And only one thread can be stopped
by entering 0, also when done repeatedly.

Always stop the old thread before (possibly) starting a new.

### DIFF
--- a/src/custom.c
+++ b/src/custom.c
@@ -744,12 +744,12 @@ static void general_option_edit(int i)
 		status_mesg(periodic_save_str, "");
 		if (updatestring(win[STA].p, &buf, 0, 1) == 0) {
 			val = atoi(buf);
-			if (val >= 0)
+			if (val >= 0) {
 				conf.periodic_save = val;
-			if (conf.periodic_save > 0)
-				io_start_psave_thread();
-			else if (conf.periodic_save == 0)
 				io_stop_psave_thread();
+				if (conf.periodic_save > 0)
+					io_start_psave_thread();
+			}
 		}
 		break;
 	case CONFIRM_QUIT:


lfos:
Sorry for the long delay. Did you make sure we never actually stop an already stopped thread using this approach? Doing that might cause a segfault on some systems. See commit 30f4112 for a possible solution.

lhca:
Not at the time I submitted the request. But since then I ran into problems with the notification thread and noticed the problem with the way threads are stopped in calcurse. It is a general problem on _all_ systems and may result in almost anything (and not only a segfault on some systems). I have just submitted a pull request #127 that has a looong explanation in the commit message.

Basically my solution is the same as that in commit https://github.com/lfos/calcurse/commit/30f411257ad3bc233184c08b846a2980a9c5d1f0: maintain a state variable that tells whether the thread is running or not. But in stead of introducing a new variable, I have used the already present thread id for that purpose. When the thread is not running, the thread id gets a (valid) value that it cannot possibly have when it is running, namely the thread id of the main calcurse program which launches all other threads. This approach has the added advantage that the thread state is correct if the `pthread_create` call fails. A fix for the periodic save thread is included.

lfos:
I merged this patch for now, thanks!

Would it make sense to use asynchronous cancellation for the save thread?

lfos:
Well, the only thing we want to prevent from is ending up with broken data files... Reading `io_save_cal()` again, I think that e55aead was not a good idea. We *should* use the mutex to guard any calls to `io_save_apts()` and `io_save_todo()`.

lhca:
Then I think the periodic save thread should abstain from using `io_save_cal()` completely - since it is heavily biased towards interactive use and has changed quite a lot in release 4.3.0. It is not suited
for periodic, automatic, non-interactive, background saving. In fact, doesn't commit 9137590 (Do not blindly overwrite files when saving) make periodic, background saves impossible?

What should the save thread as a minimum do? Just call `io_save_apts()` and `io_save_todo()` directly, possibly also `io_save_keys()`, but not `config_save()`? And how about the pre-save and post-save hooks? Is such an approach meaningful after 9137590?

lfos:
Well, for periodic save and conflicts, there is two possibilities:

1. We use some predefined (configurable?) strategy, e.g. never save if there are conflicts.
2. We do the same thing as when saving from interactive mode: Warn and display a prompt.

The second option is much more fragile since we need to take care of conflicting prompts. It is also how the current implementation works but I guess we're not careful enough here... If we switch to the first option, we should make sure we never overwrite externally modified files unless the user tells us to do so.

lhca:
A more basic problem becomes apparent: how does the periodic save thread interact with the user? It deserves a new, separate issue (which I will open in a minute).

There is too much ad hoc design here for my taste and I will do no further work on this until the basic problem is resolved.

PR #126, opened 2018-07-20
Added HTTP Support (with config option) + Fixed basic authentication

dinomacri:
1. Added configuration option to enable/disable the use of HTTPS (default is enabled), enabling support of HTTP URLs. This change simply adds a section to check whether the configuration option is enabled, and modifies the prefix accordingly.

2. Checks whether basic authentication is enabled, and correctly uses httplib2 to add the credentials to the request.

### DIFF
--- a/contrib/caldav/calcurse-caldav.py
+++ b/contrib/caldav/calcurse-caldav.py
@@ -656,6 +656,11 @@ try:
         # Authenticate with OAuth2 and authorize HTTP object
         cred = run_auth(authcode)
         conn = cred.authorize(conn)
+        
+    # Add credentials to httplib2 to be used when request requires authentication
+    if authmethod == 'basic':
+        conn.add_credentials(username, password)
+        
     elif authmethod != 'basic':
         die('Invalid option for AuthMethod in config file. Use "basic" or "oauth2"')
 
--- a/contrib/caldav/calcurse-caldav.py
+++ b/contrib/caldav/calcurse-caldav.py
@@ -533,15 +533,16 @@ try:
 except FileNotFoundError as e:
     die('Configuration file not found: {}'.format(configfn))
 
-hostname = config.get('General', 'HostName')
-path = '/' + config.get('General', 'Path').strip('/') + '/'
-hostname_uri = 'https://' + hostname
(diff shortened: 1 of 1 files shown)


lfos:
Strange. We already add the authorization headers in `get_auth_headers()`. Does you second patch actually fix a bug you discovered?

dinomacri:
First of all, this is my first contribution ever to an open source project, so I apologise if I've made any errors or misjudged anything.

My first encounter was being unable to sync with my Baikal server (http). I was receiving an ssl error (I will update with full error messages when I have access to my laptop). I dug around in the code a little and saw the "https://" being merged with the hostname. I edited this to "http://" and no longer received the error. That was my motivation for implementing the HTTP support.

My second bug (after changing to "http://") was receiving a 401 error every time I attempted a sync. I double checked the credentials I had entered in to the config file, ensured that I had uncommented "#[Auth]" along with other basic troubleshooting, but I had no success. I read through the documentation for httplib2 and I saw that "http._add_credentials_" was not used anywhere in the code. I added the line "conn.add_credentials(username, password)" and I no longer received the error, and it successfully connected.

I wasn't sure how you implemented the "get_auth_headers()" into the code, and I didn't feel qualified to change it it was best to leave it alone and present you with how I solved the issue.

Let me know if I can supply any more information, I'd be interested to know how the auth headers work, if you have the time to explain that.

Thanks.

lfos:
Maybe you could remove the patch from your script and see whether it fails again for testing purposes? Just to make sure your conclusion is correct...

lfos:
Just guessing: do you have any fancy special characters (not part of the ASCII character set) in your user name or password?

I also hope your CalDAV server is part of your local network and not some public server in the internet... Otherwise, I'd highly recommend to not use plain HTTP (which sends everything, including your login credentials without encryption),

dinomacri:
There are no fancy special characters not a part of the ASCII set, in my username or password.

lfos:
Strange. I merged the patch into `pu` for now even though I still do not understand what exactly is going on. We do set HTTP basic auth headers in `get_auth_headers()` already...

PR #127, opened 2018-07-24
Solve deadlock in notification bar

lhca:
The pull request is concerned with the way threads are stopped in calcurse. It as an alternative to commit https://github.com/lfos/calcurse/commit/30f411257ad3bc233184c08b846a2980a9c5d1f0, but also takes care of the periodic save thread and the calendar date thread.

There is a detailed explanation of the problem in the commit message.

### DIFF
--- a/src/calcurse.h
+++ b/src/calcurse.h
@@ -1208,6 +1208,7 @@ extern struct pad apad;
 extern struct nbar nbar;
 extern struct dmon_conf dmon;
 void vars_init(void);
+extern pthread_t notify_t_main, io_t_psave, ui_calendar_t_date;
 
 /* wins.c */
 extern struct window win[NBWINS];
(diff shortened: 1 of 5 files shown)


lfos:
Merged into `pu` with an additional comment added to the variables. Thanks!

lhca:
Missing thread id initialization resulted in a segmentation violation from `calcurse --status`.

lfos:
This commit is in master already, so we need to address this with a follow-up patch.

lhca:
Why is that? I pushed a simple fix to the branch on which the pull-request was based. Why not commit that to pu and master? How else do you update/change a pull-request?

lfos:
I did not see the additional patch you pushed to the branch. For some reason, the GitHub interface made it look like there's still only a single patch; sorry for that. Your fix is part of `pu`.

Issue #111, opened 2018-06-06
Configuration files in another folder


hiqua:
Issue to find a conclusion to the discussion started on #105. Read the discussion to have the full context.

In short, it would be nice to have an option `-C` to specify where the configuration files are located, especially to avoid incompatibilities when various versions of calcurse are used with the same configuration files (e.g. if they are synchronized between computers).

The question now, is what should be the defaults, and how `-C` should interact with `-D`.

@lhca @lfos let me know once you've agreed on something.

hiqua:
The conclusion would be:
* `-C $cdir` says where the configuration files should be located
* `-D $ddir`: if -C is provided, then $ddir will only contain the data files, and configuration files will be in $cdir. If -C is not provided, $ddir will contain all the files (current behaviour).
The default file layout is not changed (all files in same default directory).

Good for you @lhca  @lfos ?

lfos:
Sounds good to me! Thanks!

lhca:
Agreed.

Give it a thought whether `$cdir` should be optional with default `$ddir/config`.

PR #131, opened 2018-07-29
Use a path instead of a file for -C option

hiqua:
Allows to specify a configuration directory containing:
* conf
* keys
* hooks

When used in combination with -D $ddir, $ddir contains all the other
files not mentioned above.

I implemented it as a required argument, instead of what was suggested of having a default `config/` folder used when the argument is not given. I think this option will only be used in aliases anyway, meaning that it's ok for people to write some full path once.

Related: #111.

### DIFF
--- a/src/args.c
+++ b/src/args.c
@@ -83,7 +83,7 @@ enum {
 static void usage(void)
 {
 	printf("%s\n", _("usage: calcurse [--daemon|-F|-G|-g|-i<file>|-Q|--status|-x[<format>]]\n"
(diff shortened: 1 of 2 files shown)


lfos:
Merged to `pu`, thanks!

lhca:
There is missing a check for the existence of the alternative configuration directory (`-C <cdir>` ).

hiqua:
Thanks I'll look into it.

lfos:
This commit is in master already, so we need to address this with a follow-up patch.

PR #140, opened 2018-08-25
Check if the configuration folder exists

hiqua:
Fix an omission in #131.

@lhca can you check that I didn't forget anything this time? Thanks!

### DIFF
--- a/src/calcurse.h
+++ b/src/calcurse.h
@@ -1194,6 +1194,7 @@ extern int want_reload;
 extern const char *datefmt_str[DATE_FORMATS];
 extern int days[12];
 extern char path_dir[BUFSIZ];
+extern char path_conf_dir[BUFSIZ];
 extern char path_todo[BUFSIZ];
 extern char path_apts[BUFSIZ];
 extern char path_conf[BUFSIZ];
--- a/src/io.c
+++ b/src/io.c
@@ -231,17 +231,22 @@ unsigned io_fprintln(const char *fname, const char *fmt, ...)
 void io_init(const char *cfile, const char *datadir, const char *confdir)
 {
 	const char *home;
+	const char *conf_home;
+
 
 	if (datadir != NULL) {
 		home = datadir;
+		if (confdir == NULL)
+			conf_home = home;
+		else
+			conf_home = confdir;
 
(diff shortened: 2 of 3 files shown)


lfos:
We should probably add tests for the new flag. Can be done in a follow-up series, though. Merged into `pu`. Thanks!

lhca:
@hiqua Well, you asked for it! There are also checks for directories in `args.c()` for non-interactive mode.

Maybe @lfos could see to that, and at the same time also fix `path_hooks` which is missing in both interactive and non-interactive mode.

hiqua:
I'll have a look (at least for the `arcs.c()` thing), thanks!

lhca:
There are problems with this patch, and they are not solved by https://github.com/lfos/calcurse/commit/e3ca60eefe08b253ff436fe16d85b4317c7a2523.

It may be a copy-paste typo, but the three original lines

> snprintf(path_conf, BUFSIZ, "%s/" CONF_PATH_NAME, conf_home);
snprintf(path_keys, BUFSIZ, "%s/" KEYS_PATH_NAME, conf_home);
snprintf(path_hooks, BUFSIZ, "%s/" HOOKS_DIR_NAME, conf_home);

works when -C confdir is given, but not when -C is absent.

With the three new lines it is the other way around:

> snprintf(path_conf, BUFSIZ, "%s/" CONF_PATH, conf_home);
snprintf(path_keys, BUFSIZ, "%s/" KEYS_PATH, conf_home);
snprintf(path_hooks, BUFSIZ, "%s/" HOOKS_DIR, conf_home);

The reason is that the macros ending in _PATH and _DIR include `.calcurse/` and are for the default case only, while those ending in _NAME are file names and are for the non-default cases. The use of them must be conditioned by `confdir`.

The `io_init()` routine can only be read with great effort and should be rewritten. There should also be a check on the size of the user supplied directory paths. They are copied into fix-sized buffers. I have attached a suggestion.
[io_init.txt](https://github.com/lfos/calcurse/files/2592263/io_init.txt)

lfos:
True. I'd be happy about the patch. We should avoid using buffer sizes and magic constants, though. And if possbile, we should even avoid snprintf() and use the string_*() functions or asprintf() instead. Thanks!

PR #141, opened 2018-08-30
Save load review

lhca:
Release 4.3.0 of calcurse included several features to support save and reload of data while taking changed data files into account.

I have reviewed the commits from a year ago and the result is this pull request. There is nothing really new here, but several bug fixes and a reshuffling of the deck. I believe the result is a simplication of the save/reload routines and an improvement of the user interface.

I will quickly go through the commits. The first four are just preparations, clearing the way for what follows. The next two rework two of the last year's commits; in particular 57dd3d6 was flawed (bugs are still present i 4.3.0). The rest are (hopefully) improvements.

There are two main ideas/principles that make simplifications possible/easier in the save/reload routines:

1) Let `io_load_data()` initialize lists, check for modified files and decide what to load.
2) Move curses-dependent user info and window updates to the command level in calcurse.c

For further details I refer to the commit messages.


src/io.c
```diff
@@ -532,25 +532,37 @@ static int resolve_save_conflict(void)
...
 }
 
-static int io_check_data_files_modified()
+/* Return codes for new_data(). */
+#define NONEW		0
+#define APTS		1
+#define TODO		2
+#define APTS_TODO	3
+#define NOKNOW		4
+static int new_data()
```
lfos: Can't we simply use `TYPE_MASK_CAL` and `TYPE_MASK_TODO` here and return a negative value (i.e. -1)  on error?

lhca: Yes, it should be made clearer that the return value is a bit mask. However, it is purely internal to io.c, and the bit masks in args.c may change for other reasons. A return value of -1 does not go well with a bit mask: it's nice that `io_load_data()` can just check bit 1 and 2 and pass on the return value from `new_data()`.

I will upload a correction that also has an improved error check in `io_reload_data()`.

lfos: Well there are two possibilities: if we want to return a value that corresponds to the empty bit mask, we should return 0. If we want to say "there is an error and we cannot return a valid bit mask", I'd say that -1 is a good way to signal this.

If, for some reason, you actually want to use another bit in the mask to mark an error (which only really makes sense if the other bits could still be either set or unset in the error state), you should make that clear in the code and check for errors using `new_data() & MASK_ERROR`. I also recommend dropping `APTS_TODO` and defining `APTS` as `1 << 0`, `TODO` as `1 << 1` etc.

lhca: Agreed.

lfos:
Merged to `pu`, thanks!

PR #148, opened 2018-09-17
Note export functionality (DESCRIPTION) for ical + newline fix for unix systems

DavidSpenler:
I added the function `ical_format_newlines` to convert LF line endings to CR+LF, which is the standard for the ical format. This is called after all vevent components are written to the file. I changed the file stream `stream`'s method from `w` to `w+` to facilitate this.

I also added functionality to export the note of appointments and events, and todos using the DESCRIPTION parameter of the ical standard, as it is currently unused and is roughly analogous to the note parameter of calcurse's. To accomplish this, I added the function `read_note_content` to `note.c`, as well as its prototype to `calcurse.h,` which extracts the note content using the filename stored in the respective structure.

### DIFF
--- a/src/calcurse.h
+++ b/src/calcurse.h
@@ -955,6 +955,7 @@ void edit_note(char **, const char *);
 void view_note(const char *, const char *);
 void erase_note(char **);
 void note_read(char *, FILE *);
+char* read_note_content(char *);
 void note_gc(void);
 
 /* notify.c */
--- a/src/ical.c
+++ b/src/ical.c
@@ -97,6 +97,56 @@ static void ical_export_footer(FILE * stream)
 	fputs("END:VCALENDAR\n", stream);
 }
 
+static void ical_format_newlines(FILE * stream) 
+{
+	char *buffer;
+	long length,newlines = 0;
+	int c,i = 0;
+
+	fseek(stream,0,SEEK_SET);
+
+	while (c != EOF) {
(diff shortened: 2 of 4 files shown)


lfos:
Thanks for working on this! In general, I like the idea. When it comes to the implementation, there are several things to be addressed, though.

Firstly, we should not mix different features/fixes in the same commit. Please address the newlines issue and the notes export in separate commits. Also, for the newlines stuff, I do not see why we would need another pass replacing the LF? Why can't we use the proper delimiters when printing stuff? In addition to the fact that using a second pass is clunky and unnecessary, I think switching from `w` to `w+` would also break the ability to export to stdout, right? That feature is needed in many places, such as for CalDAV synchronization.

DavidSpenler:
Thanks for looking over my changes. I'll make separate pull requests, but first I'll address your concerns.

The icalendars specifications requires that .ical files use CR+LF newlines (https://icalendar.org/iCalendar-RFC-5545/3-1-content-lines.html)

As such, files that use only LF can not be successfully imported into popular calendar applications such as Google Calendar.

After some reading I found that we can disable newline translating if `fopen` is called with the `wb` instead. I'll make the changes later today as well as the new pull requests.

lfos:
Yes, I am fine with `CR+LF` but it should be done in a single pass. Looking forward to the new PRs!

Issue #145, opened 2018-09-12
Calcurse doesn't redraw while adding a task


NWuensche:
Hello,

I'm using Calcurse v 4.3.0 on Arch Linux and I have a problem when I change the size of the terminal while I'm adding an entry. Normally, Calcurse redraws when I change the terminal size with i3wm or tmux. However, when I'm adding a task (pressing `a`), Calcurse doesn't redraw properly. I use terminator 1.91 as a terminal emulator and i3 in version 4.15 and tmux 2.7.

Normal behavior when not adding a task when resizing fullscreen terminal to half-size:
![12 09 2018 17 14 11](https://user-images.githubusercontent.com/15856197/45434802-8d3e4300-b6af-11e8-8e5f-debf80437514.png)

Behavior when currently adding a task:
![12 09 2018 17 13 56](https://user-images.githubusercontent.com/15856197/45434749-70097480-b6af-11e8-90dc-f86b5ef6a13b.png)

lfos:
Yes, it's a known bug. It's useful to have it recorded on the tracker, though; thanks!

PR #150, opened 2018-09-26
Cancel editing in getstring() on window resize

lhca:
Adresses #145. It seems a reasonable compromise that the edit is aborted.

### DIFF
--- a/contrib/caldav/Makefile.am
+++ b/contrib/caldav/Makefile.am
@@ -4,7 +4,8 @@ dist_bin_SCRIPTS = \
 	calcurse-caldav
 
 EXTRA_DIST = \
-	calcurse-caldav.py
+	calcurse-caldav.py \
+	config.sample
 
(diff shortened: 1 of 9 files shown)


lfos:
I am not convinced that aborting is a better idea than having the panels not updated for a while... While the latter is not esthetically pleasing, the former means that data is potentially lost.

lhca:
The input generated by a resize should be taken care of in any case. Another possibility is to handle `KEY_RESIZE` like `ERR` and ignore it (i.e. `continue;`) and hope that the messed-up screen allows the editing to continue.

lfos:
True. I'd prefer the option to ignore the input as a workaround; maybe with a "FIXME" comment that mentions that resize should be handled properly at some point. Thanks!

lfos:
Pushed an alternative workaround, thanks for the idea!

Issue #134, opened 2018-07-31
Periodic save and user interaction


lhca:
This is a continuation of a discussion in #114. I quote @lfos:

> Well, for periodic save and conflicts, there is two possibilities:
  We use some predefined (configurable?) strategy, e.g. never save if there are conflicts.
  We do the same thing as when saving from interactive mode: Warn and display a prompt.

> The second option is much more fragile since we need to take care of conflicting prompts. It is also how the current implementation works but I guess we're not careful enough here... If we switch to the first option, we should make sure we never overwrite externally modified files unless the user tells us to do so

Both options involve interaction with the user: displaying messages as well as receiving input. But neither is possible from the periodic save thread because the main calcurse program owns and runs the screen (apart from the notification bar, more below). The problem is not new, but is becoming more apparent as the save function `io_save_cal()` is evolving. To witness, perform the following experiment:

- start calcurse and open the general options configuration menu; set periodic save to 3 (minutes, so you don't have to wait too long); then go back to the main screen
- with calcurse still running, open the `conf` file in the `.calcurse` directory with an editor and insert a line at the top containing a single `x` (any syntax error will do); write the file and leave the editor
- wait for the next periodic save to happen

An "INTERNAL ERROR" box appears out of nowhere, and the clock stops running in the notification bar. The box has been opened by the periodic save thread which is now waiting for user confirmation (any key). It has locked a mutex that is also used by the notification thread, hence the stopped clock. The main calcurse program is also waiting for user input in its main loop. It is anyone's guess which of the two will get the user input (curses has only one input queue). If the save thread, it will exit still owning the locked mutex and leaving the error box behind, if the main loop it will depend on the character typed (the screen may be (partly) redrawn). And calcurse remains locked.

Calcurse is not designed to cope with a "background" thread interacting with the user like this. For the thread running the notification bar special care has been taken. The thread has its own one-line window where output is displayed, and it does not receive user input.

lfos:
It is true that we should be more careful here. However, I do not see how the first option (using a fixed non-interactive strategy for conflicts) requires any user interaction from the background thread. Could you please elaborate?

For the second option, a proper fix probably involves notifying the main thread that the save operation has failed and performing any kind of user interaction from there (which is possibily delayed until after any ongoing user interactions are completed; we might even add some manual delay between user interaction and the prompt to avoid accidental inputs from the user, otherwise they might send keystrokes not meant to be a reaction to the prompt).

lhca:
[...]

PR #153, opened 2018-10-03
System queue

lhca:
This is an implementation of a system message queue as outlined in #134.

I have used already existing tools in calcurse: linked lists and the appointment/event mechanism. I refer to the commit messages for further explanations. The pull request is based on #141.


lfos:
Merged to `pu`, thanks!

PR #158, opened 2018-11-01
DST updates

lhca:
The recent end of Daylight Saving Time on 28 October caused some deficiensies to be exposed.

One type had to do with the length of a day, another with increments in dates (adding a time interval to a point in time). A straightforward bug has also been fixed. Each type is in a seperate commit.

In most of the cases I have reproduced the error before the correction and verified its absence afterwards.


lhca:
I forgot: in most cases I have reproduced the error before the correction and verified its absence afterwards.

lfos:
Great! Merged into pu, thanks!

PR #161, opened 2018-11-14
fix for not importing prioritys correctly

dottdottdott:
Just a small fix to enable imporitng priorities from ical files


dottdottdott:
The testfiles also need to be fixed. The icals got todo items with priority 1 and the scripts check for 0.

PR #162, opened 2018-11-21
Cli update

lhca:
An updated man page, calcurse(1).

The man-page is mostly concerned with the CLI, and going through it led to some CLI bug fixes as well. There are also two interface changes:

- the day range for queries (`--from`, `--to`) is now inclusive; previously the end day, `--to`, was not included
- the input date format is taken from the configuration file, but may be overridden by a new long option `--input-datefmt`; time (of day) input has been eliminated, but has apparently never been used

The last commit (Safety exit and read-only mode) is put here out of convenience. It has an update and fix for https://github.com/lfos/calcurse/commit/2339d78cbd562319473a34a36390378df64ade00 and is not related to CLI or the other commits.

### DIFF
--- a/doc/calcurse.1.txt
+++ b/doc/calcurse.1.txt
@@ -30,519 +30,733 @@
  */
 ////
 
-calcurse(1)
+CALCURSE(1)
 ===========
 
-Name
+NAME
 ----
 
-calcurse - text-based organizer
+calcurse - customizable calendar app with reminders and command line features
 
-Synopsis
+SYNOPSIS
 --------
 
-[verse]
-'calcurse' *-Q* [options] [*--from* <date>] [*--to* <date>|*--days* <num>]
-'calcurse' *-G* [options]
-'calcurse' *-i*<file>
-'calcurse' *-x*<format>
-'calcurse' *--gc*
-'calcurse' *--status*
-'calcurse' *--version*
-'calcurse' *--help*
+--
+*calcurse* [-D 'directory'] [-C 'confdir'] [-c 'calendar-file']
 
-Description
+*calcurse* -Q [*--from* 'date'] [*--to* 'date' | *--days* 'number']
+
+*calcurse* -a | -d 'date' | -d 'number' | -n | -r['number'] | -s['date'] | -t['number']
+
+*calcurse*  -G | -P ['filter options'] ['format options']
+
+*calcurse* -h | --status |  -g | -i 'file'  | -x['file'] | --daemon
+--
+
+The first form shows how to invoke calcurse interactively; the remainder is
+command line forms.
+
+The second form shows queries (as opposed to interactive use). For
+convenience, common queries have abbriviated forms shown in the third line.
+All queries may be combined with filter options as well as format options.
+
+The fourth form shows operations on the calcurse data files, one for
+display of entries, the other for removal of them.
+
+The fifth form is miscellaneous help and support functions.
+
+All details are in <<_options,OPTIONS>>.
+
+DESCRIPTION
 -----------
 
-Calcurse is a text-based calendar and scheduling application. It helps keeping
-track of events, appointments and everyday tasks.  A configurable notification
-system reminds user of upcoming deadlines, and the curses based interface can
-be customized to suit user needs.  All of the commands are documented within an
-online help system.  
+calcurse is a calendar and scheduling application for use in a terminal
+session (terminal emulator). When invoked without options, calcurse enters
+interactive mode; in most other cases, calcurse produces output on the
+terminal and exits.  It helps keeping track of events, appointments and
+everyday tasks. Interactive mode is used when data are entered or when already
+existing entries are inspected or edited. All data are saved to disk as text
+files. Command line mode is used for queries and administrative tasks and for
+automating tasks in scripts.
 
-Options
+The interactive interface is based on ncurses and can be customized to suit
+user preferences. Customization includes program behaviour as well as visual
+appearance and key bindings, and is performed interactively; the result is
+automatically saved to disk and loaded on every invocation. Available actions
+are documented in an online help system. A configurable notification system
+issues reminders of upcoming deadlines.
+
+When leaving the interactive program, a background daemon may continue running
+and issue reminders; it stops automatically when the interactive mode is
+reentered.
+
+This man page mainly describes the command-line mode. The following two
+subsectiions contain some general desriptions of command line options and usage.
+
+Input and Output Date Format
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Many options require a 'date' argument, and query results per day are set apart 
+by a leading 'date line'.
+
+The input format of 'date' options and the output format of 'date lines' are
+taken from the configuration file (see <<_files,FILES>>). The formats are set
+in the "General Options" submenu in interactive mode.  Particularly in scripts
+it may be desirable that formats are independent of the local user
+configuration. For this purpose use the options *--input-datefmt* and
+*--output-datefmt*.
+
+An input date consists of 'date', 'month' and 'year'. Here 'day' must be in
+the range 1-31 and 'month' in 1-12. Depending on the operating system 'year'
+must be in the range 1902-2037 or 1902-?.  Additionally, some short forms are
+allowed with the obvious meaning: +today+, +tomorrow+, +yesterday+, +now+ and
+weekdays +mon+, ..., +sun+.
+
+Optionally, a 'date' argument for a filter option (see
+<<_filter_options,Filter Options>>) may be followed by a time-of-day
+specification in hours and minutes (24-hour clock). The specification has the
+fixed format hh:mm (example: +"2018-12-1 20:30"+ when the input date format is
+the ISO standard format). Note that the entire date specification must be
+quoted to create one argument.
+
+Filter, format and day range options
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+These options do not accomplish anything in and of themselves. They influence
+other options and are in a way opposites: filter options affect the input to,
+format and day range options the output from calcurse. Specifically, filter
+options limit what is loaded from the data files into calcurse, day range
+options limit what is output (see <<_query,-Q>>), and
+<<_format_options,format options>> describe how it is presented.
+
+Filter options have effect on queries (*-Q* and query short-forms), grep
+(*-G*), purge (*-P*) and export (*-x*). Format options have effect on queries,
+grep and *--dump-imported*. Day range options have effect on queries only.
+
+OPTIONS
 -------
 
-The following options are supported:
+Most options imply command line mode. Options compatible with interactive mode
+are marked "('also interactively')".
 
 *-a*, *--appointment*::
-  Print the appointments and events for the current day and exit. Equivalent to
-  *-Q --filter-type cal*. 'Note:' The calendar from which to read the
-  appointments can be specified using the *-c* flag.
+  Print the appointments and events for the current day. Equivalent to *-Q
+  --filter-type cal*.
 
-*-c* <file>, *--calendar* <file>::
-  Specify the calendar file to use. The default calendar is *~/.calcurse/apts*
-  (see section 'FILES' below). This option has precedence over *-D*.
+*-c* 'file', *--calendar* 'file'::
+  ('also interactively') Specify the calendar file to use. The default
+  calendar is *~/.calcurse/apts* (see <<_files,FILES>>). If 'file' is not an
+  absolute path name, it is interpreted relative to the current working
+  directory. The option has precedence over *-D*.
 
-*-d* <date|num>, *--day* <date|num>::
-  Print the appointments for the given date  or  for  the given  number  of
-  upcoming days, depending on the argument format. Two possible formats are
-  supported:
+*-C* 'dir', *--conf* 'dir'::
+  ('also interactively') Specify the configuration directory to use. If not
+  specified, the default directory is *~/.calcurse/*. See <<_files,FILES>> for
+  the interaction with *-D*.
+
+*-D* 'dir', *--directory* 'dir'::
+  ('also interactively') Specify the (data) directory to use. If not specified,
+  the default directory is *~/.calcurse/*.  See section <<_files,FILES>> for
+  the interaction with *-C*.
+
+*-d* 'date|num', *--day* 'date|num'::
+  Print appointments and events for the given date or given range of days,
+  depending on the argument format:
 +
 --
-  * a date (possible formats described below).
-  * a number *n*.
+  * a 'date'
+  * a number 'num'
 --
 +
-In the first  case,  the  appointment  list  for  the specified  date will be
-returned, while in the second case the appointment list for the *n*  upcoming
-days will be returned.
-+
-As an example, typing *calcurse -d  3*  will  display your  appointments  for
-today, tomorrow, and the day after tomorrow. The first form is equivalent to
-*-Q --filter-type cal --from <date>*, the second form is equivalent to *-Q
---filter-type cal --days <num>*.
-+
-'Note:' as for the *-a* flag, the calendar  from which to  read  the
-appointments can be specified using the *-c* flag.
+In the first case, appointments and events for 'date' are returned, while in
+the second case those for 'num' days are returned.  Positive values of 'num'
+means the future, negative the past; the range either starts or ends with the
+current day. As an example +calcurse -d 3+ displays appointments and events
+for today, tomorrow and the day after tomorrow, while +calcurse -d -2+
+displays those for yesterday and today. The first form is equivalent to *-Q
+--filter-type cal --from* 'date', the second to *-Q --filter-type cal --days*
+'num'.
 
 *--daemon*::
-  Start calcurse in background mode. Restart if the daemon was already running.
+  Start calcurse in background mode; restart, if the daemon was already
+  running. Usually done automatically by setting the configuration option
+  +daemon.enable+ in the 'Notify' submenu in interactive mode.
 
-*--days* <num>::
-  Specify the length of the range (in days) when used with *-Q*. Cannot be
-  combined with *--to*.
+*--days* 'num'::
+  Specify the range of days when used with *-Q*. Can be combined with
+  *--from*, but not with *--to*. Without *--from*, the first day of the range
+  defaults to the current day. The number may be negative, see
+  <<_query,-Q --query>>.
+
+*--dump-imported*::
+  When importing items, print each newly created object to stdout.  Format
+  options can be used to specify which details are printed. See also
+  <<_format_options,Format Options>>.
 
 *--export-uid*::
   When exporting items, add the hash of each item to the exported object as an
   UID property.
 
-*-D* <dir>, *--directory* <dir>::
-  Specify the data directory to use. If not specified, the default directory is
-  *~/.calcurse/*.
+*--from* 'date'::
+  Specify the start date of the day range when used with *-Q*. When used
+  without *-to* or *--days* the range is one day (the specified day), see
+  <<_query,-Q --query>>.
 
 *-F*, *--filter*::
-  Read items from the data files and write them back. The filter interface can
-  be used to drop specific items. Be careful with this option, specifying a
-  wrong filter might result it data loss! It is highly recommended to test the
-  effect of filters with -G first. See also: 'Filter Options'.
-
-*--from* <date>::
-  Specify the start date of the range when used with *-Q*.
+  Deprecated, see <<_purge,-P, --purge>>. Note that this option is for
+  backward compatibility and not the same as *-P* (it does not use the invert
+  filter option.
 
 *-g*, *--gc*::
-  Run the garbage collector for note files and exit.
+  Run the garbage collector for note files. The garbage collector removes
+  files from the +notes+ directory (see <<_files,FILES>>) that are no longer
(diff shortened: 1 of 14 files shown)


lfos:
I guess we should also update the test suite and enforce the new documented behavior?

lhca:
There were only two failed test and they have been updated (https://github.com/lfos/calcurse/pull/162/commits/91930248fd8a0197b297d9530355ec74e583e10e) so that the command line input format agrees with the `conf` file.

And I had to change the test configuration file as well (and set `general.systemdialogs=yes`) even though previous Travis builds succeeded without it. How can that be?

I tried to figure out a way to make the proposed `calcurse(1)` available in HTML form, but couldn't find one (I don't have a home page).

lfos:
I had a closer look at the patches and they mostly look good to me. I am not sure about the patch dropping time ranges. We should allow for, say, listing all appointments starting from 3 p.m. to 10 a.m. on the next day.

The `general.systemdialogs` thing is weird and we should definitely debug that instead of simply turning it on. Maybe it is because something is now unexpectedly printed in CLI mode with system dialogs turned on.

What I meant in the previous comment is that we should also extend the test cases and add tests for the new "inclusive date range" behavior.

You can build an HTML version of the man page by using asciidoc directly on the text file.

lhca:
> I had a closer look at the patches and they mostly look good to me. I am not sure about the patch dropping time ranges. We should allow for, say, listing all appointments starting from 3 p.m. to 10 a.m. on the next day.

It's true that a time of day is no longer accepted as part of a date, but nothing was dropped. What you wish may be desirable, but it is not possible in 4.3.0.
 
> The `general.systemdialogs` thing is weird and we should definitely debug that instead of simply turning it on. Maybe it is because something is now unexpectedly printed in CLI mode with system dialogs turned on.

It's the other way around. The seven ical tests expect an import report like

> Import process report: 0017 lines read
1 app / 0 events / 1 todo / 0 skipped

This is only output when systemdialogs is on, and all ical tests failed when I ran the tests. The Travis build still succeeds, so apparently it succeeds, whether systemdialogs is on or off. I have no way of debugging that. The function `show_dialogs()` is a wrapper for `conf.system_dialogs` (=`general.systemdialogs`), and used to control whether save-messages were displayed. By intention the status bar messages about save and reload cannot be suppressed, so import reports are the only output controlled by `general.systemdialogs`
> 
> What I meant in the previous comment is that we should also extend the test cases and add tests for the new "inclusive date range" behavior.

Definitely.
 
> You can build an HTML version of the man page by using asciidoc directly on the text file.

That was not the problem I had in mind. The Makefile has a target for that. But I could find no way of _publishing_ the result for comments.

lhca:
> > I had a closer look at the patches and they mostly look good to me. I am not sure about the patch dropping time ranges. We should allow for, say, listing all appointments starting from 3 p.m. to 10 a.m. on the next day.
> 
> It's true that a time of day is no longer accepted as part of a date, but nothing was dropped. What you wish may be desirable, but it is not possible in 4.3.0.

After thinking more about this: you are right, and what you ask for _is_ possible in 4.3.0 (with the proper filter options). My confusion comes from the words "time ranges". I prefer "time of day" (Uhrzeit).

Let me try to clarify my own thoughts. Command line input dates are used in two different contexts:

- to specify a day range for output from queries (`--from`, `--to`, `--days`)
- to specify start and end times in filter options (`--filter-start-from`, etc.)

In the first context a time of day should be ignored and only the date (Datum) taken into account. In the second context the date with an optional time of day should be accepted.

Is this meaningful? Do you agree?

lfos:
IIRC, there is an important difference between the filter options and the query range options. The filters are applied when the items are loaded. They can only be used to filter based on item properties but not on occurrence properties. For example, `--filter-start-from` would not give you any future occurrences of a recurrent item if the first occurrence of that item is before the filter date.

The query range options can be used to filter occurrences. So it would be nice to have "time of day" support for them as well.

It's been quite a while since I have been working on this, so you might want to double-check.

lhca:
Reintroducing time of day in the input date format is of course easy enough. Supporting it in both contexts mentioned above is another matter.

The filter option context will require different handling of date only and date with time of day (because the first is a time span of 24 hours, the second a point in time). But that should be doable, and is then an enhancement of behaviour already present. I will work on that.

As for the query range context see below.

> IIRC, there is an important difference between the filter options and the query range options. The filters are applied when the items are loaded. They can only be used to filter based on item properties but not on occurrence properties. For example, `--filter-start-from` would not give you any future occurrences of a recurrent item if the first occurrence of that item is before the filter date.

Thank you. That's also my understanding. The new man-page tries to make the difference clear with a similar example (in the box "Start and end time of events and recurring items"). 

> The query range options can be used to filter occurrences. So it would be nice to have "time of day" support for them as well.

The man-page example similarly mentions that the query range may be used to limit the output. But requiring that it should work not just for dates, but also for dates with time of day (granularity of seconds) is new. This is definitely _not_ possible in 4.3.0, so there is no question of anything being dropped here.

Support for this is a major effort in my estimation. The input date is used directly to loop through the query range by increments of a day, and is passed on as argument to the `day_store_*()` functions. I can't assess whether it's possible or meaningful at all.

lfos:
Ok, I'd say updating the man page is fine then. We can improve this in a future version of calcurse. Will have a closer look at all the patches soon. Thanks for all the work!

lfos:
Did not review in full detail but merged everything into `pu`. Will have a closer look at the code and man page changes soon. Looks very promising, thanks for working on this!

lfos:
I decided to reopen this until it is merged into master. An HTML version of the (current version of the) new man page can be found under [1]. I will review that and comment here.

[1] https://calcurse.org/files/calcurse.1-pu.html

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
+remaining are command line forms. The second form shows queries, the third
+short-form queries; both may be combined with filter options as well as
+format options. The fourth form is miscellaneous help and support functions.
+All details are in <<_options,OPTIONS>>.
 
-Options
+DESCRIPTION
+-----------
+
+Calcurse is a calendar and scheduling application for use in a terminal
```
lfos: I know that this has just been copied from the previous revision of the file but I wonder whether we should always use lower case for the proper name `calcurse`, even if it is at the beginning of a sentence. Opinions?

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
+reentered.
+
+This man-page mainly describes the command-line mode.
+
+OPTIONS
 -------
 
-The following options are supported:
+.Input and Output Date Format
+*****************************
```
lfos: I'd skip that heading. It may look nice and consistent in the generated HTML but the primary purpose of this document is to serve as a man page and the additional heading is very uncommon there.

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
+  Display *calcurse* version.
+
+*-x*['format'], *--export*[='format']::
+  Export user data in the specified format. Events, appointments and todos are
+  converted and echoed to stdout. Two formats are available: +ical+ and
+  +pcal+. The default 'format' is +ical+.
+
+[[_limiting_options]]
+.Filter, format and day range options
+*************************************
```
lfos: Again, please make this a proper level-1 heading as it was before. While the new structure is semantically "correct", we should focus on making things easily visible in the man page. We should also avoid having many subheadings which make the man page harder to read.

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
+times and apply to appointments and events only, and the 'todo filters' concern
+priority and completeness and apply to TODOs only.
+
+Outside the three groups is the invert filter.
+
+*--filter-invert* ::
+  Invert the combined effect of any other filters, i.e. load the items that
+  do 'not' match them.
+
+.'general filters'
```
lfos: Those subheadings in particular look very strange in the new man page. The formatting of the `general filters` heading is different from the `calendar filters` heading.

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
-
-Format specifiers for appointments
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-*s*::
-  Print the start time of the appointment as UNIX time stamp
-*S*::
-  Print the start time of the appointment using the *hh:mm* format
-*d*::
+[horizontal]
```
lfos: These commands generate labelled tables in the generated man page which look quite odd...

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
-save before quitting).  All of these options are detailed in the configuration
-menu.
+`calcurse -G --filter-type apt --format-apt "%(start:%c) %(duration:%d %H:%M)\t%m\n"`::
+  For each appointment list the start time in a localized standard format,
+  followed by the duration in days, hours and minutes, followed by the
+  description.
 
-Files
+[[_files]]
+FILES
```
lfos: From a cursory read, I really like the updates to the files section and everything below. I will have a closer look soon!

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
+appearance and key bindings, and is performed interactively; the result is
+automatically saved to disk and loaded on every invocation. Available actions
+are documented in an online help system. A configurable notification system
+issues reminders of upcoming deadlines.
+
+When leaving the interactive program, a background daemon may continue running
+and issue notifications; it stops automatically when the interactive mode is
+reentered.
+
+This man-page mainly describes the command-line mode.
```
lfos: "man page", no hyphen.

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
-'calcurse' *-x*<format>
-'calcurse' *--gc*
-'calcurse' *--status*
-'calcurse' *--version*
-'calcurse' *--help*
-
-Description
------------
+--
+*calcurse* [-D 'directory'] | [-C 'confdir'] | [-c 'calendar-file']
```
lfos: `datadir` instead of `confdir`?

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
+This man-page mainly describes the command-line mode.
+
+OPTIONS
 -------
 
-The following options are supported:
+.Input and Output Date Format
+*****************************
+Many options require a 'date' argument, and query results per day are set apart 
+by a leading 'date line'.
```
lfos: I see what you intended with this introductory section but I am not sure whether it is all that useful. A man page is used like a dictionary most of the time and rarely read from cover to cover. Maybe we should move all this to the formatting section? As a user, I'd expect to find it there.

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
+other options and are in a way opposites: filter options affect the input to,
+format and day range options the output from calcurse. Specifically, filter
+options limit what is loaded from the data files into calcurse, day range
+options limit what is output (see <<_query,-Q>>), and
+<<_format_options,format options>> describe how it is presented.
+
+Filter options have effect on queries (*-Q* and query short-forms), grep
+(*-G*), purge (*-P*) and export (*-x*). Format options have effect on queries,
+grep and *--dump-imported*. Day range options have effect on queries only.
+*************************************
```
lfos: Again, I do not think this is very useful in practice. Let's rather turn all subsections into proper sections and move the pieces of information from this box to the intros of the corresponding new sections (where not present already).

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
+[[_extended_format_specifiers]]
+.Extended format specifiers
+
+Extended format specifiers can be used to control the printing of times for
+some of the single-letter specifiers. Additionally there are two specifiers
+that do not have any corresponding short form and are intended for use in
+scripting.
+
+*%d* ~ *%(duration*`[`*:*'format'`]`*)*::
+*%r* ~ *%(remaining*`[`*:*'format'`]`*)*::
```
lfos: I see you intentation here, but even as a calcurse developer I was confused by this syntax. Can we simply drop the short versions here and add something like "%(duration) is the long version of %d and %(remaining) is the long version of %r" instead?

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
+  *%S* with the following modifications: 1) days are not limited to the
+  "calendar" values 0-31 (hours, minutes and seconds are "clock" values, but
+  see *E* in the following) 2) each number is by default padded with zeros to
+  two decimal places, 3) the *%* character may be followed by one or two
+  optional flags: *-*, which suppresses the zero-padding, *E*, which will
+  suppress the "clock" limits on *%H*, *%M* and *%S*; if both are used, *-*
+  must precede *E*, 4) no other flags or width specifications are supported
+
+*%s* ~ *%(start*`[`*:*'format'`]`*)*::
+*%e* ~ *%(end*`[`*:*'format'`]`*)*::
```
lfos: Same here. I think we can simply drop `*%s* ~` and `*%e* ~` here since they are already mentioned in the text below.

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
+  description.
 
-Files
+[[_files]]
+FILES
 -----
 
-The following structure is created in your $HOME directory (or in the directory
-you specified with the *-D* option), the first time calcurse is run:
+The following structure is created by default in your $HOME directory the
```
lfos: Maybe we can change "$HOME directory" to "home directory" while touching these lines...

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
-directory. This file contains logs about calcurse activity when running in
-background.
+Calcurse may switch between two configuration setups, but still access
+the same data files e.g. with:
 
-Environment
+----
+$ calcurse
+
+$ calcurse -C $HOME/.calcurse/config
```
lfos: This should be quoted: `calcurse -C "$HOME/.calcurse/config"`.

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
-*post-save*::
-  Executed after the data files are saved.
-
-Some examples can be found in the `contrib/hooks/` directory of the calcurse
-source tree.
+If the calcurse data files contain many entries which are no longer needed or
+wanted, they can, of course, be removed interactively. If there are many, it
+can be a tedious task and may be done better as in the following two examples.
+'Note.' The filter options specify what is retained, not what is removed!
+Presently it is not possible to list the items to be removed.
```
lfos: Is this true? Can't we use `-G --filter-invert`?

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
-calcurse(1)
+CALCURSE(1)
 ===========
 
-Name
+NAME
 ----
 
-calcurse - text-based organizer
+calcurse - terminal-based organizer for interactive and command line use
```
lfos: Should we come up with a good concise description to use everywhere? Right now, the man page, documentation, the web site and distribution packages each use their own descriptions.

It should make the core features obvious, include the most relevant search keywords (calendar, scheduling, command line) and be as short as possible...

doc/calcurse.1.txt
```diff
@@ -30,519 +30,714 @@
...
------------
+--
+*calcurse* [-D 'directory'] | [-C 'confdir'] | [-c 'calendar-file']
+
+*calcurse* -Q [*--from* 'date'] [*--to* 'date' | *--days* 'number']
+
+*calcurse* -a | -d 'date' | -d 'number' | -n | -r['number'] | -s['date'] | -t['number']
+
+*calcurse* -h | --status | -G | -g | -i 'file' | -P | -x['file'] | --daemon
+--
```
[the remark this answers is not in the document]
lhca: Synopsis section.

The first line is obviously wrong. Corrected.

I don't agree as to the other lines. The query option (`-Q`) stands apart. It is the command line equivalent of the interactive interface and is the only option for which the day range concept is applicable and definitely not on a par with `-P` and `-G`. The third line lists all the various query short forms.

The last line lists the remaining options.

It is an attempt to give a quick overview of a mixed bag of options.

doc/calcurse.1.txt
lhca: Corrected.

doc/calcurse.1.txt
lhca: I admit that I mainly tried to make the HTML version look nice at the expense of the traditional terminal version. I have long since lost contact with the day-to-day IT world. Do (Unix) people still read man pages with the man command? Anyway, I have gone back to a traditional layout with just a few section levels (three).

doc/calcurse.1.txt
lhca: Corrected.

doc/calcurse.1.txt
lhca: Corrected.

doc/calcurse.1.txt
lhca: Corrected.

doc/calcurse.1.txt
lhca: Corrected.

doc/calcurse.1.txt
lhca: ?? I considered `datadir` instead of `directory`.

doc/calcurse.1.txt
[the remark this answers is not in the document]
lfos: I agree that `-Q` is special. However, these lines also help increase the visibility of the different modes of operation. I don't feel strongly about it but I'd personally prefer having `-G` and `-P` grouped in a separate line for that reason. If that does not convince you, we can keep it like that.

doc/calcurse.1.txt
lfos: Yes, I believe that people still rely on man pages, especially when looking for a specific command line option. Of course, Google often helps as well but I'd say man pages are still being used, especially among heavy shell users (which is one of the main target groups of calcurse).

Beyond that, this is the source code for the man page. If we decide that we want to drop the man page and replace it with HTML documentation, we should do that instead of keeping the man page and making it ugly, while trying to make it convert into a pretty HTML page .

doc/calcurse.1.txt
lfos: Yes, that's what I meant, sorry.

doc/calcurse.1.txt
lfos: If the home directory contains any space, the command without the quotes breaks. It's a corner case but it does not hurt to convert the example into a proper command that works under all circumstances.

doc/calcurse.1.txt
lfos: I thought of this before and the only thing I do not like about it is that it's too long. I think we should aim for something with <80 characters but I cannot think of anything better right now.

lfos:
Could you also try to rebase this branch on master (or pu) while reworking it? There only is a small conflict in `args.c`. That would make it much easier for me to merge and test stuff. I will give you some time to discuss and incorporate the suggested changes and later build and upload a new HTML version of the suggested man page.

Thanks!

lhca:
> Could you also try to rebase this branch on master (or pu) while reworking it? There only is a small conflict in `args.c`.

Ah, yes. I didn't notice that. It's the `io_init()` rewrite overtaking the CLI-update into pu. I have pushed an update.

lfos:
The io_init() patch has been merged to master already, so we cannot amend that. This branch needs to be rebased on master instead.

lfos:
While interactively rebasing, you could also squash the man page review updates into the original man page patch :)

doc/calcurse.1.txt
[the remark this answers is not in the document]
lhca: They have got their own line.

doc/calcurse.1.txt
lhca: I retained _directory_, because it contains both data and configuration files when `-D` is used alone.

doc/calcurse.1.txt
lhca: Corrected. Spaces crept into directory names a long time ago. Sigh.

doc/calcurse.1.txt
lhca: <80 is a tall order. This one is 77 characters long:

calcurse - customizable calendar app with reminders and command line features

Here 'interactive use'  is understood as opposed to  'command line'. I believe that 'app' is in common use for 'program' or 'application'.

lhca:
> While interactively rebasing, you could also squash the man page review updates into the original man page patch :)

I'm still not quite at ease with rebasing. Here is what I have done. I removed the `io_init()` stuff from the CLI-update branch and squashed all the man page commits into the original patch. Then I forced-pushed everything to GitHub without rebasing. Will that do?

lfos:
Thanks! I merged everything to pu with minor modifications, rebuilt and uploaded the HTML version. I like the outcome. Will merge to master soon unless there are any objections.

doc/calcurse.1.txt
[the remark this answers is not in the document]
lhca: I think that your suggestion is an improvement, but it's not in pu. Didn't you like it?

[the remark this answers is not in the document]
lhca: > I think that your suggestion is an improvement, but it's not in pu. Didn't you like it?

My mistake. I just realized that the last round of changes didn't make it into my local CLI-update branch before the squashing and pushing (grr and some danish swearing).

lfos:
I added most (but not all) of your updates to pu. Thanks!

lfos:
Merged into master.

PR #163, opened 2018-12-03
[fix] debug was breaking auth

Psycojoker:
Hello,

Objects are reference in python, ".pop" was modifying the original dictionary and thus breaking authentication by removing auth headers.

Kind regards,


lfos:
Merged. Thanks!

PR #164, opened 2018-12-04
Rewrite of io_init()

lhca:
Mainly to improve readability. The statically allocated buffers can be replaced in a followup patch.


src/io.c
```diff
@@ -140,67 +140,45 @@ unsigned io_fprintln(const char *fname, const char *fmt, ...)
...
-
-		snprintf(path_todo, BUFSIZ, "%s/" TODO_PATH, home);
-		snprintf(path_cpid, BUFSIZ, "%s/" CPID_PATH, home);
-		snprintf(path_dpid, BUFSIZ, "%s/" DPID_PATH, home);
-		snprintf(path_notes, BUFSIZ, "%s/" NOTES_DIR, home);
-		snprintf(path_dmon_log, BUFSIZ, "%s/" DLOG_PATH, home);
-	}
+#define SNPRINTF(d, s1, s2) \
+  EXIT_IF(snprintf(d, BUFSIZ - 20, "%s%s", s1, s2) >= BUFSIZ - 20, \
+          _("Directory name too long: %s (limit %d)"), s1, BUFSIZ - 21);
```
lfos: This is a very non-performance critical data path. Can we simply use asprintf() to avoid these magic numbers and potential pitfalls with buffer sizes?

lhca: Done in new commit.

lfos:
I had a quick look at the new implementation and I like it a lot better. The macro is still slightly confusing, though. For a couple of seconds I was very confused about how the code can work without a format string until I had a closer look. Can't we simply use asprintf() directly and add the mem_free() manually to the only line where it is actually needed?

lhca:
I don't understand:

> Can't we simply use asprintf() directly and add the mem_free() manually to the only line where it is actually needed?

Only line? Isn't it needed for every path?

In this case you could also simply decide to ignore the freeing of allocated memory. In the default case `io_init()` is only called once, and when `-D`, `-C` or `-c` is present, it is called at most three times. So after all, at most a few hundred bytes are leaked once.

lfos:
Well, io_init() is only called once and all the variables default to NULL. I do not see why we should need to free them. I pushed an alternative patch to pu, let me know what you think.

lhca:
Fine with me. Please ignore my previous comment. My own calcurse version has an io_init() call in main() and I managed to mix things up.

lhca:
Just for the record.
 
> Well, io_init() is only called once ...

I have pushed a commit to CLI-update so that this is still the case.

lhca:
> Just for the record.
> 
> > Well, io_init() is only called once ...
> 
> I have pushed a commit to CLI-update so that this is still the case.

There was a mistake on my part concerning how may times `io_init()` is called or should be called in `parse_args()`, and the commit mentioned above corrected it so that _all_ datafile options are parsed first. It's not in pu, so it seems I never did as promised. But now I have.

PR #167, opened 2019-01-07
Explicit view configuration

lhca:
Entries for calendarview and todoview in the general options menu.


lfos:
Merged into pu, thanks!

PR #175, opened 2019-01-26
Deletion of an occurrence of a multiday recurrent appointment

lhca:


lfos:
Good catch, thanks. I also realized that we do not seem to be resetting the calendar cache when removing an appointment. If you have an occurrence of a recurrent appointment as the only item on a selected day and remove that occurrence, the day is still marked as busy in the calendar panel.

lhca:
Yes, indeed.

And another detail: you may have noticed that the variable `date` in `ui_day_item_delete()` at the end is passed on to `day_cut_item()`. The change to `date` in the pull request has no unintended effects here, because `date` is not used by `day_cut_item()` and should really be removed.

Issue #178, opened 2019-02-04
Export command on curses interface doesn't save before closing interface


crvs:
pressing 'x' and providing an exported filename produces a 0 bytes file until closing the application

PR #179, opened 2019-02-04
closing the export stream after being done with export

crvs:
solves lfos/calcurse#178

close the stream when done writing to it in `io_export_data` this results in the export file being updated immediately.


lfos:
Merged with a whitespace fix, thanks!

PR #183, opened 2019-02-14
Fix empty dynamic string

lhca:
No known bug fixed by this, but during work on a new feature I ended up editing non-initialized memory.


lfos:
Merged, thanks!

PR #184, opened 2019-02-14
Exception days

lhca:
When you delete an occurrence of a recurrent event or appointment (as opposed to the entire repeated item), the day is added internally as an exception day.

This works well, but subsequently there is no way of removing the exception day (and get the occurrence back). In fact, given a repeating event/appointment, you cannot easily determine whether some occurrences have been deleted. You can, of course, view or even edit the data file `apts` (but that is cheating).

This commit makes it possible to edit the list of exception days by extending `(4) Repetition` of the edit item command. If exception days have been added, they are presented in the status line as a string of space-separated dates in your preferred input format. It may be viewed or edited as you like. The intention is that you only view or remove days. But there is nothing to stop you from adding new exception days, and they will do no harm. But there is no check that they are meaningful, only that they are valid dates.

### DIFF
--- a/src/calcurse.h
+++ b/src/calcurse.h
@@ -1008,6 +1008,8 @@ void pcal_export_data(FILE *);
 /* recur.c */
 extern llist_ts_t recur_alist_p;
 extern llist_t recur_elist;
+void recur_update_exc(llist_t *, char *);
+char *recur_exc2str(llist_t *);
 struct recur_event *recur_event_dup(struct recur_event *);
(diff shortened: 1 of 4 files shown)


lfos:
I really like this. Merged! I think editing exception days can be cumbersome if there really are a lot of dates. I'd be happy to have some option to edit those days in an external editor. Maybe we should add a general key binding to the prompt that puts the current `getstring` line in a temporary file, lets you edit the line, and re-reads the line into the prompt once the editor is closed.

Issue #187, opened 2019-02-25
[Request] Add custom directory support to calcurse-caldav


vxid:
This is probably not really high priority but it would be nice to add a data directory flag to calcurse-caldav to complement the -D flag in calcurse to indicate which directory events should be saved in.
If i'm not mistaken, calcurse-caldav currently hardcode `~/.calcurse` as the storage path which causes synchronization to fail if events are in a different directory.

lfos:
Huh? Using `-D`, you should be able to change the directory that events are saved in...

vxid:
I do not think I exposed my problem clearly.

I currently use a different directory, `$HOME/usr/cal`, to store calendar objects.

Using `calcurse -D $HOME/usr/cal -C $HOME/usr/cal` allows me to use data files from this directory as expected (the config files also are in the same directory, hence the `-C` flag).

However, syncing using ```calcurse-caldav --config $HOME/usr/cal/caldav/config --lockfile $HOME/usr/cal/caldav/lock --syncdb $HOME/usr/cal/caldav/sync.db``` still saves remote events to `$HOME/.calcurse` and assumes local events are also present in this directory.

I'm started working on a PR to add the -D flag as a config option to the `calcurse-caldav` script, it seems like it wouldn't take much effort to pass this new -D option to subprocess calls from within the script.

lfos:
Do you still plan on working on this for `calcurse-caldav` as well? Sounds like a good idea to me.

vxid:
Yes, I will open a PR once calcurse-vdirsyncer is merged.

Issue #35, opened 2017-06-04
vdir support ?


strk:
Any plan in supporting the VDIR storage format for calendar events ?
See https://vdirsyncer.pimutils.org/en/stable/vdir.html

lfos:
Not yet. Patches are welcome, though!

guygma:
I am also interested in this. Any sense of the kind of effort it would take at this point? It may provide more standards compliant caldav compatibility than the current contribution.

lfos:
Well, I guess you would at least have to rewrite the calcurse-caldav script [1]. You can have a look to get an idea of what needs to be done.

[1] https://github.com/lfos/calcurse/blob/master/contrib/caldav/calcurse-caldav.py

guygma:
That is one approach, but to me that seems like it would unnecessarily reinvent what vdirsyncer already does well. The problem of talking to a CalDAV server and maintaining a bi-directional synchronized copy of its contents is already solved. I had envisioned a translation between the vdir format and the internal format of calcurse (most simple); or, better yet, a direct implementation of reading from vdir text format instead of using the calcurse default backend.

lfos:
True, importing and exporting from/to vdir sounds like a good idea. Reading and writing directly from/to vdir should be possible as well. If we decide to go that route, we should probably support that for iCal as well.

guygma:
I could be mistaken on this, but I do not think there is much of a distinction made for iCal format -- at least if all you mean by that is files with the ".ics" extention. For vdirs, contacts are stored in ".vcf" and calendars are stored in ".ics" as per the standards of vCard and iCal, respectively. I presume todos are included in the iCal representation but I won't swear to that.

lfos:
Mh, I have not looked into vdir yet. If that's really all that needs to be done, we might "support" vdir already, though. We do support iCal imports and exports.

guygma:
Yeah, give it a look ([vdirsyncer](https://github.com/pimutils/vdirsyncer)). It is a very nice project. It sounds like it could be a straightforward implementation to streamline full support of vdirs, which inherently makes calcurse as compatible with CalDAV as vdirsyncer has implemented (which is quite complete).

[...]

PR #188, opened 2019-02-27
[WIP] vdirsyncer support

vxid:
I wrote a < 100 loc script to sync calcurse with a remote using vdirsyncer (see #35) which works already really well since vdirsyncer does all the heavy lifting.

So far the script does the following:

1. Export all calcurse items and their uids using the same method as found in `calcurse-caldav`.
2. Save the exported items in the vdir directory and delete old events not present in calcurse anymore.
3. Sync to remote using vdirsyncer.
4. Remove elements delete from remote in calcurse, and import newly created ones

The script works without any issues when syncing with a Fastmail remote.

I am planning on improving the script in the next few days by:
- adding configuration flags (see #187) and removing hardcoded paths 
- Adding synchronization filters (e.g. evt, todo).

Once these improvements are made, this script would pretty much have the same features as `calcurse-caldav` since vdirsyncer takes care of the entire syncing process.

@lfos What other functionality or requirements would you consider necessary before potentially merging this PR?

### DIFF
--- a/contrib/vdirsyncer/calcurse-vdirsyncer.py
+++ b/contrib/vdirsyncer/calcurse-vdirsyncer.py
@@ -1,0 +1,181 @@
+#!/usr/bin/env python3
+
+import argparse
+import io
+import os
+import re
+import subprocess
+import sys
+import textwrap
+
+
+def msgfmt(msg, prefix=''):
+    """Print a formatted message"""
+    lines = []
+    for line in msg.splitlines():
+        lines += textwrap.wrap(line, 80 - len(prefix))
+    return '\n'.join([prefix + line for line in lines])
+
+
+def check_binary(binary):
+    """Check if a binary is available in $PATH"""
+    try:
+        subprocess.call([binary, '--version'], stdout=subprocess.DEVNULL)
+    except FileNotFoundError:
+        sys.exit(msgfmt("{0} is not available.".format(binary), "error: "))
+
+
+def check_directory(directory):
+    """Check if a directory exists"""
+    if not os.path.isdir(directory):
+        sys.exit(msgfmt("invalid directory: {0}".format(directory), "error: "))
+
+
+def write_file(file, contents):
+    """Write to file"""
+    if verbose:
+        msgfmt("Writing to file {0}".format(file))
+    with open(file, 'w') as f:
+        f.write(contents)
+
+
+def remove_file(file):
+    """Remove file"""
+    if verbose:
+        msgfmt("Deleting file {0}".format(file))
+    if os.path.isfile(file):
+        os.remove(file)
+
+
+def calcurse_export():
+    """Return raw calcurse data"""
+    command = calcurse + ['-xical', '--export-uid']
+    proc = subprocess.Popen(command, stdout=subprocess.PIPE)
+    return [x for x in io.TextIOWrapper(proc.stdout, encoding="utf-8")]
+
+
+def calcurse_remove(uid):
+    """Remove calcurse event by uid"""
+    command = calcurse + ['-P', '--filter-hash=' + uid]
+    subprocess.call(command)
+
+
+def calcurse_import(file):
+    """Import ics file to calcurse"""
+    command = calcurse + ['-i', file]
+    subprocess.call(command)
+
+
+def calcurse_list():
+    """Return all calcurse item uids"""
+    command = calcurse + [
+        '-G',
+        '--format-apt=%(hash)\\n',
+        '--format-recur-apt=%(hash)\\n',
+        '--format-event=%(hash)\\n',
+        '--format-recur-event=%(hash)\\n',
+        '--format-todo=%(hash)\\n'
+    ]
+    proc = subprocess.Popen(command, stdout=subprocess.PIPE)
+    return [x.strip() for x in io.TextIOWrapper(proc.stdout, encoding="utf-8")]
+
+
+def vdirsyncer_sync():
+    """Sync vdir using vdirsyncer"""
+    command = vdirsyncer + ['sync']
+    if force:
+        command += ['--force-delete']
+    subprocess.call(command)
+
+
+def parse_calcurse_data(raw):
+    """Parse raw calcurse data to a uid/ical dictionary"""
+
+    header = ''.join(raw[:3])
+    regex = '(BEGIN:(VEVENT|VTODO).*?END:(VEVENT|VTODO).)'
+    events = [x[0] for x in re.findall(regex, ''.join(raw), re.DOTALL)]
+
+    items = {}
+
+    for item in events:
+        uid = re.findall('UID:(.*?)\n', item)[0]
+        items[uid] = header + item + "END:VCALENDAR\n"
+
+    return items
+
+
+def calcurse_to_vdir():
+    """Export calcurse data to vdir"""
+    raw_events = calcurse_export()
+    events = parse_calcurse_data(raw_events)
+
+    files_old = [x for x in os.listdir(vdir)]
+    files_new = [uid + ".ics" for uid in events]
+
+    if not files_new == 0 and not force:
+        sys.exit(msgfmt("""no calcurse events were exported."""
+                        """ Use --force to synchronize anyways.""", 'error: '))
+
+    for file in files_old:
+        if file not in files_new:
+            remove_file(os.path.join(vdir, file))
+
+    for uid, event in events.items():
+        file = uid + ".ics"
+        if file not in files_old:
+            write_file(os.path.join(vdir, file), event)
+
+
+def vdir_to_calcurse():
+    """Import vdir data to calcurse"""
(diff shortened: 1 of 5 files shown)


lfos:
Neat! Your plan sounds good to me. Maybe try to follow PEP8 (you can use Pylint or a similar tool to check for potential issues).

Do you plan to maintain this script and fix possible bugs in the future?

vxid:
Thank you! I did use pylint but I silenced a few errors, I will refactor the script today.
I don't mind maintaining the script at all and I plan on using it daily anyways.

lfos:
I think we can merge a first version once all the basic things are is configurable.

Please make sure you add a nice commit message quickly explaining the functionaility and usage of the script as well. A simple README similar to [1] would be nice as well.

Anything else can be added later.

Thank you so much for working on this!

[1] https://github.com/lfos/calcurse/blob/master/contrib/caldav/README.md

vxid:
My branch is ready. I also changed the makefiles to include the script.

contrib/vdirsyncer/calcurse-vdirsyncer.py
```diff
@@ -0,0 +1,150 @@
...
+def vdirsyncer_sync():
+    """Sync vdir using vdirsyncer"""
+    command = vdirsyncer + ['sync', '--force-delete']
+    subprocess.call(command)
+
+
+def parse_calcurse_data(raw):
+    """Parse raw calcurse data to a uid/ical dictionary"""
+    header = ''.join(raw[:3])
+    events = re.findall('(BEGIN:.*?END:.*?.\n)', ''.join(raw[3:]), re.DOTALL)
```
lfos: This looks a bit fragile. Can we at least turn `BEGIN:` into something like `BEGIN:(VEVENT|VTODO)` and also only match at the beginning of lines?

I am also not too convinced to rely on the fact that the header is three lines long and then hardcode the "footer" at the same time. But maybe that's good enough for a first PoC.

contrib/vdirsyncer/calcurse-vdirsyncer.py
```diff
@@ -0,0 +1,150 @@
...
+    files_old = [vdir + x + '.ics' for x in calcurse_list()]
+    files_new = [vdir + x for x in os.listdir(vdir)]
+
+    for file in files_new:
+        if file not in files_old:
+            calcurse_import(file)
+
+    for file in files_old:
+        if file not in files_new:
+            calcurse_remove(file.replace(vdir, "").replace(".ics", ""))
```
lfos: Again, I am sure there is a better way to do this. Why don't you drop the `vdir` prefix from `files_old` and `files_new`? You can then simply write `calcurse_import(vdir + file)` above and do not need to do the first string replacement here.

Maybe we can also turn `for x in os.listdir(vdir)` above into `for x in os.listdir(vdir) if x.endswith('.ics')`? Just to be safe.

And maybe use `[:-4]` instead of `.replace(".ics", "")`.

contrib/vdirsyncer/readme.md
```diff
@@ -0,0 +1,69 @@
+calcurse-vdirsyncer
```
lfos: Can we rename this file to `README.md`? That's what we use everywhere in this project.

contrib/vdirsyncer/readme.md
```diff
@@ -0,0 +1,69 @@
...
+
+- Baikal
+- DavMail (Exchange, Outlook)
+- FastMail
+- Google
+- iCloud
+- nextCloud
+- ownCloud
+- Radicale
+- Xandikos
```
lfos: I am not sure whether it is helpful to describe vdirsyncer features here, especially since it means we will have to maintain and update this list. But it probably does not hurt too much either...

contrib/vdirsyncer/readme.md
```diff
@@ -0,0 +1,69 @@
...
+
+Planned Updates
+---------------
+
+- Support for hook directories
+- Enable filtering of imported and exported items (events, todos)
```
lfos: Great, really looking forward to this.

Thanks for working on this!

contrib/vdirsyncer/calcurse-vdirsyncer.py
```diff
@@ -0,0 +1,150 @@
...
+
+def check_directory(directory):
+    """Check if a directory exists"""
+    if not os.path.isdir(directory):
+        sys.exit("Error. Invalid directory {0}".format(directory))
+
+
+def write_file(file, contents):
+    """Write to file"""
+    print("Writing to file {0}".format(file))
```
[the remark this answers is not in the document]
vxid: I added a verbose flag

contrib/vdirsyncer/calcurse-vdirsyncer.py
```diff
@@ -0,0 +1,150 @@
...
+    if os.path.isfile(file):
+        print("Deleting file {0}".format(file))
+        os.remove(file)
+
+
+def calcurse_export():
+    """Return raw calcurse data"""
+    command = calcurse + ['-xical', '--export-uid']
+    proc = subprocess.Popen(command, stdout=subprocess.PIPE)
+    return [x for x in io.TextIOWrapper(proc.stdout, encoding="utf-8")]
```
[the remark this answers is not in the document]
vxid: It is the same in this case, but in some other cases I strip the individual elements in the same operation (for instance in `calcurse-list`). I used the same notation here to stay consistent.

contrib/vdirsyncer/calcurse-vdirsyncer.py
```diff
@@ -0,0 +1,150 @@
...
+    command = calcurse + ['-P', '--filter-hash=' + uid]
+    subprocess.call(command)
+
+
+def calcurse_import(file):
+    """Import ics file to calcurse"""
+    command = calcurse + ['-i', file]
+    subprocess.call(command)
+
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
vxid: Oops. Probably didn't see it after merging/rebasing just before pushing. I did test the script multiple times, but not before pushing. Removed.

contrib/vdirsyncer/calcurse-vdirsyncer.py
vxid: I agree that this part is not robust.

It is actually not necessary to keep the header at all, but it is handy since it automatically tracks the correct calcurse version number in the ical `PRODID:` field (hence no need to add it manually afterwards, using a potentially wrong version number). However, assuming that the header is exactly 3 lines is risky indeed.

As for the regex itself, I changed the pattern to include (VTODO|VEVENT) and it seems to work well enough (I am no regex expert at all though). I do not know how to match line beginnings only since newlines are counted as raw characters.

A bonus (or bug, I really do not know) of using the new regex is that it automatically extracts the event type as well, which means it will be easier to filter event types in the future.

contrib/vdirsyncer/calcurse-vdirsyncer.py
vxid: Done, thank you for the suggestions.

vxid:
I corrected the issues you pointed out and added a verbose flag to the script.
Do you think the script is good enough for a merge now?

lfos:
Looks good to me. Two small things:

1. Now that `files_old` and `files_new` only contain file names (no full path) in `vdir_to_calcurse()`, can we do the same in `calcurse_to_vdir()` for consistency (i.e. only add `vdir` to the `remove_file()` and `write_file()` calls)?
2. Can you make sure you test the *final* version of the script properly before submitting? I don't have time to setup vdirsyncer and do proper testing, and I will trust your assertion that everything has been tested.

vxid:
The changes are made.

I did test the script multiple times (including an import of all my events on another machine), and everything seems to work fine. I also tested the build with the new Makefile and `calcurse-vdirsyncer` is available in `$PATH` afterwards.

lfos:
One more thing I noticed: does the script actually work if you pass `vdir` without a trailing slash? It definitely should. I would recommend using `os.path.join(vdir, file)` instead of string concatenation.

vxid:
Good call!

If no slash is supplied, the script fails silently (no syncing occurs, meaning that all items created after the last sync in calcurse are deleted).

I fixed the issue as you suggested.
Just to be on the safe side, I also added a new flag and two checks during the sync process:

- If no events are exported when calling calcurse-export, the script issues an error and stops in order to prevent the deletion all items in the vdir directory. This can be overridden by using the new `--force` flag if this is intended.
- If the vdir directory is empty after the synchronization with vdirsyncer (probably meaning that all items were delete on the remote), a similar error message is issued in order to prevent the deletion of all local calcurse items. This can also be overridden using the `--force` flag.
- The `--force` flag also implicitly enables the `--force-delete` option in vdirsyncer, allowing to delete the remote collection is all items are deleted locally. That was a default so far but I think it makes more sense to turn it off by default.

lfos:
Merged into pu, thanks a lot for working on this!

PR #190, opened 2019-03-03
Multiple days in the appointments panel

lhca:
This has been WIP for a couple of months, and the commit history reflects the fact that ideas were introduced and later dropped.

Design considerations  are in the commit log messages, particularly

- Multiple days in the appointments panel
- Redesign of selected-item implementation for the APP panel
- Multiple days: selected day derived from selected item

There are three new user configuration settings in general options:

- multiple days: number of days (1..21) to display in the appointments panel, default seven
- dayseparator: lines between days in the appointments panel, default two
- emptyline: insert an empty line after each appointment, default true

The first one should be obvious; set it so that the days (on average) fill the entire panel. There is nothing gained from loading more days than can be displayed. Up/down movement in the panel works across days with new days loaded as needed. Insert, delete, edit etc work for the selected day/item. The selected day in the calendar changes with movements and other operations in the panel.

The second one can be set to 0, 1 or 2 to separate the days clearly or squeeze as many as possible into the panel.

The third one is for backward compatibility: appointments used to consist of three lines, the last one always empty to separate it from the following appointment. You can leave the empty line out and squeeze even more lines into the panel.

There are some changes in the calendar as well. The week number used to be the week of the day in the appointments panel. With multiple days it makes more sense to display week numbers for the entire month in a column to the left. Also, dates are shown for overlapping parts of the first and last week of the month. The day number (in the year) of the selected day has replaced the old week number.

At first I thought that the range of days currently loaded should be indicated in the calendar with brackets (`[...]`). This turned out to be not useful and a bit confusing and was dropped. What remained, though, was the square brackets to indicate the currently selected day. This has the additional benefit compared to the previously highlighted date that it does not obscure the colour of the date.

The selected day can still be explicitly set in the calendar, and all move commands work as before (`t, T, w, W, ...`), but as mentioned above the selected day is now automatically set from operations in the appointments panel. In this way the selected day, which is tied to the calendar, plays a less prominent part than before.


lfos:
Wow, this is a massive patch set. First of all, thank you so much for putting all the effort into implementing this!

Is there any chance we can turn this into a reviewable patch series, though? I imagine something like a handful of commits without any fixups. Also, there seem to be some conflicts that need to be addressed.

lhca:
> Is there any chance we can turn this into a reviewable patch series, though? I imagine something like a handful of commits without any fixups.

Please explain. I don't understand.

> Also, there seem to be some conflicts that need to be addressed.

I'll look into those. Mostly from work in parallel, I  believe.

lfos:
> > Is there any chance we can turn this into a reviewable patch series, though? I imagine something like a handful of commits without any fixups.
> 
> Please explain. I don't understand.

Before submitting a PR to a project, it is pretty common to refactor the patch set such that it looks like a clean series of atomic with clear history. This can be achieved by using a feature like `git rebase -i`, squashing some patches, and doing a bit of code refactoring.

It does not only make the review process *much* easier but also avoids unnecessarily cluttering the project's commit history (thereby making it harder for future developers to understand the development of a feature or tracking down bugs using a feature like `git bisect`).

Let me know if you think it is too much work. If it is, maybe we can at least squash some of the patches.

lhca:
That's perfectly clear and only reasonable. My conclusion is that this commit series is not ready for a PR. But there is no hurry. After looking at the conflicts I can also see that some cleanup is needed anyway.

How do you withdraw a PR? Just close it?

lfos:
We can close it. On the other hand, we can simply leave it open until everything is ready.

lhca:
The new PR #193 contains 18 commits of which at least five are minor. As I write this, I realize that three fixups could have been squashed as well.

PR #191, opened 2019-03-04
Fix vector_remove()

lhca:
The function was temporarily used in the multiple-days implementation.


lfos:
Merged into pu, thanks!

PR #192, opened 2019-03-05
Add custom calcurse data directory support to calcurse-caldav

vxid:
See issue #187

Added the --datadir flag to calcurse-caldav which allows to specify a custom data directory similarly to the -D flag in calcurse.

I had to change the `calcurse` datatype from string to list to make this work cleanly.

### DIFF
--- a/contrib/caldav/calcurse-caldav.py
+++ b/contrib/caldav/calcurse-caldav.py
@@ -36,6 +36,11 @@ def die(msg):
     sys.exit(msgfmt(msg, "error: "))
 
 
+def check_dir(dir):
+    if not os.path.isdir(dir):
+        die("invalid directory: {0}".format(dir))
+
+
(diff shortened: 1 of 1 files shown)


lfos:
This looks good to me. One thing that bothers me a little  is the inconsistency between the calcurse option (`--directory`) and the calcurse-caldav option (`--datadir`). I must say that I prefer the calcurse-caldav naming, though. Maybe we can rename the calcurse option to `--datadir` as well? And rename `--conf` to `--confdir` at the same time?

Merged to pu.

vxid:
Sounds good to me. Do you want to do this in a separate PR?

lfos:
You're more than welcome to do this in a separate PR. Make sure to update the man page and documentation as well. And we should keep the old options for backwards compatibility (but do not need to mention them in the man page anymore).

Closing this PR for now.

Issue #152, opened 2018-10-02
Feature Request: Show more than one day of appointments


FoolishExcalibur:
I was wondering if there was a possibility to show more than one day worth of appointments in the app. window? Like you wouldn't need to go to the next day to see what kind of appointment it is but rather have a bunch listed from the beginning?

lfos:
This is not currently possible. Sounds like a useful feature, though.

FoolishExcalibur:
The only currently to me known method to do so is by appending -d*insert high number* 
It shouldnt be all too difficult to implement it.

Ierezell:
@FoolishExcalibur Yes right now you can do it in CLI but I was thinking the same as you....
Please, could it be possible to list all the appointement in the appointment window (or at least all for the month or week) ? 

Btw, great job it's really a nice program to use !

lhca:
I am looking into this and have a prototype mostly working that diplays a week at a time.

Anyone willing to try it out and provide feedback? I will make it available as a branch in the cloned project https://github.com/lhca/calcurse. You will have to build from source. We can discuss which base to use.

koreymacdougall:
@lhca - I'd be interested in trying this out.  I looked quickly through your branches but couldn't find the commit/changes.  Is it currently up?

lhca:
@koreymacdougall - It's not there yet because it's not quite ready. When it is, I will have to port it from my own private branch (not public) to one that is accessible, either pu or master? Have you any preferences?

koreymacdougall:
@lhca - it wouldn't make any difference to me which branch it's on.  I'd be testing it as an isolated source build, away from a system install, so it wouldn't conflict with the mainline master, if that's what you're thinking about.  

I'm really interested to see your implementation.  I've been meaning to write this patch myself for awhile, but haven't had the time.  Glad to see someone making progress on it!

lhca:
@koreymacdougall - Here it is https://github.com/lhca/calcurse/tree/multiple-days (based on pu).

Basically, seven days are loaded into the appointments panel instead of just one. Hopefully, it's intuitively clear how to proceed from there.

[...]

PR #193, opened 2019-03-05
Multiple days in the appointments panel

lhca:
This has been WIP for a couple of months, but has been cleaned up and rearranged for this PR. The base branch is therefore not the previous one of the same name.

Design considerations are in the commit log messages, particularly

- Multiple days in the appointments panel
- Multiple days: Redesign of selected-item implementation for the APP panel
- Multiple days: calendar display
- Multiple days: selected day derived from selected item

There are three new user configuration settings in general options:

- multipledays: number of days (1..21) to display in the appointments panel, default seven
- dayseparator: lines between days in the appointments panel, default two
- emptyline: insert an empty line after each appointment, default true

The first one should be obvious; set it so that the days (on average) fill the entire panel. There is nothing gained from loading more days than can be displayed. Up/down movement in the panel works across days with new days loaded as needed. Insert, delete, edit etc work for the selected day/item. The selected day in the calendar changes with movements and other operations in the panel.

The second one can be set to 0, 1 or 2 to separate the days clearly or squeeze as many as possible into the panel.

The third one is for backward compatibility: appointments used to consist of three lines, the last one always empty to separate it from the following appointment. You can leave the empty line out and squeeze even more lines into the panel.

There are some changes in the calendar as well. The week number used to be the week of the day in the appointments panel. With multiple days it makes more sense to display week numbers for the entire month in a column to the left. Also, dates are shown for overlapping parts of the first and last week of the month. The day number (in the year) of the selected day has replaced the old week number.

At first I thought that the range of days currently loaded should be indicated in the calendar with brackets ([...]). This turned out to be not useful and a bit confusing and was dropped. What remained, though, was the square brackets to indicate the currently selected day. This has the additional benefit compared to the previously highlighted date that it does not obscure the colour of the date.

The selected day can still be explicitly set in the calendar, and all move commands work as before (t, T, w, W, ...), but as mentioned above the selected day is now automatically set from operations in the appointments panel. In this way the selected day, which is tied to the calendar, plays a less prominent part than before.

Hopefully it is intuitively clear how to proceed - it is still the old calcurse look-and-feel.


lhca:
I have further covered my tracks by squashing three fixups into an existing commit thus reducing the number of commits to 15.

Two commits are not directly concerned with multiple days, but adress issues in the basic scrollwindow/listbox code.

src/calcurse.c
```diff
@@ -48,18 +48,13 @@ int count, reg;
  */
 static void do_storage(int day_changed)
 {
+	/* * Save the selected item before rebuilding the day vector. */
```
lfos: Does this extra `*` have any purpose? Also, I see why you moved that comment there but it is now less prominent than it was before.

src/utils.c
```diff
@@ -672,6 +672,12 @@ time_t get_today(void)
 	return date2sec(day, 0, 0);
 }
 
+/* Returns the beginning of the selected day in the calendar. */
+time_t get_slctd_day(void)
+{
+	return date2sec(*ui_calendar_get_slctd_day(), 0, 0);
+}
+
```
lfos: Not sure whether I like that. The idea was to have all functions directly accessing UI properties named with a `ui_` prefix. I agree that `ui_calendar_get_slctd_day_sec()` might not have been the best name but the new one is inconsistent with the remaining naming. I also believe it belongs to some of the UI units and not to the general utility functions...

Talking about naming: is there any reason the subject line of this commit starts with `Multiple days`? I don't see how this is related to the multiple days feature. You might have done it as a preparatory cleanup to that series but I think it might cause confusion instead of adding value for future readers of the commit history...

src/ui-day.c
```diff
@@ -693,9 +682,11 @@ void ui_day_item_add(void)
 			item.ev = event_new(item_mesg, 0L, start, 1);
 		}
 		io_set_modified();
-		day_store_items(get_slctd_day(), 1, day_get_nb());
-		ui_day_load_items();
-		ui_day_set_selitem_by_aptev_ptr(item);
+		/* Set the selected APP item. */
+		struct day_item d = empty_day;
+		d.order = start;
+		d.item =item;
```
lfos: Nit: missing whitespace.

src/day.c
```diff
@@ -734,26 +734,32 @@ int day_paste_item(struct day_item *p, time_t date)
...
+		p = VECTOR_NTH(&day_items, i);
+		/* Same day and item? */
+		if (p->order == order && p->item.apt == item.apt)
+			return i;
+	}
+	VECTOR_FOREACH(&day_items, i) {
+		p = VECTOR_NTH(&day_items, i);
+		/* Same item? */
+		if (p->item.apt == item.apt)
+			return i;
```
lfos: I found this confusing and reading the commit message or the comments did not help. What is the purpose of the new `order` argument? What does this double search do? Without reading the remaining code in detail: I guess the first loop tries to find an occurrence on the current day? But then, what does the other loop do? What exactly is the (new?) purpose of this function?

Also, again I think we can skip the `Multiple days` from the commit message. It makes it overly long and does not add any value.

lfos:
There seem to be a couple of failing tests on this PR. Did you already check whether they are actually related to the changes introduced by this patch set?

src/calcurse.c
lhca: This was corrected by commit https://github.com/lfos/calcurse/pull/193/commits/7a137e98a8696eb9dd11cfe756a2cd996de3085b.

src/day.c
lhca: Re `order`: see commit log message https://github.com/lfos/calcurse/pull/193/commits/be8a363208ab7fc09644c628e43255b6eeb167ed.

src/utils.c
lhca: I see your point about `ui_`. I treated the selected day like the current day. After all, the two variables

`static struct date today, slctd_day;`

live side by side in `ui-calendar.c`. Would you prefer `ui_get_slctd_day()`? in `ui_calendar.c`?

lhca:
The failing tests _are_ related to the PR. I am working on them.

General comments to your reviews above.

- I _did_ make an effort to make everything ready for review, but do you expect every code line to remain unchanged once introduced?
- the GitHub interface is not very helpful in showing which commit is being reviewed
- the `Multiple days` in the subject line of the commits is simply to show that they were submitted as a whole; a few were inserted as part of the cleanup; I'll remove them (and any other) again if you like

lfos:
> * I _did_ make an effort to make everything ready for review, but do you expect every code line to remain unchanged once introduced?

That would be optimal but not strictly necessary. A compromise is fine if the additional effort in cleaning up the last couple of lines clearly exceeds the benefits.

> * the `Multiple days` in the subject line of the commits is simply to show that they were submitted as a whole; a few were inserted as part of the cleanup; I'll remove them (and any other) again if you like

The purpose of commit messages is not (only) to explain the series to a reviewer of the patch series. The main purpose is to help a future reader of the commit history understand why the change was made.

I did also test the new UI in practice and have a couple of comments that I will share later (either today or after the upcoming weekend).

lfos:
My promised first impressions on the UI experience: it is certainly great to be able to view more than a day. I have a very large display with a rather small font and I usually have enough space to display ~1 week of appointments in the panel.

Sometimes, I found the new appointment and calendar panels quite confusing and often found it hard to keep track of which day I am currently viewing. Now, the fact that I am used to the "old" interface after using it for many years certainly contributes to that confusion. But still, even when trying to be very objective, I found the new interface much less clear. Do not take this as me rejecting the idea: I like it in general and we will probably have compromise simplicity when adding powerful features to the UI.

I think one of the factors might be that the separation between different days is not clear enough. The vertical space is configurable but that might not be enough. For example, there is a horizontal separator (line) if you have events and appointments on the same day. For me, that line is a much stronger visual separation than a space of, say, three lines. Thus, if you have multiple days with mixed events and appointments, the visual separation hierarchy is not in line with the actual logical hierarchy. Give it a try and let me know whether or not you agree.

Another point is that the current day might be in the middle of the screen. I would probably find it easier to keep track of the currently selected day if it were always on top, even though that would mean less flexibility and remove the possibility the view all events of the past week.

Let me know what you think.

lhca:
In general, I don't share your concerns, but then I am biased and would have done something about them if I had. You address two issues: separation of the days in the appointments panel, and keeping track of the selected day (for clarity: by "current day" I understand "today").

Both bring up a basic principle of the calcurse user interface, namely the use of highlighted text to guide the user's attention. The calcurse screen has three graphical levels: background, fixed foreground text (parts of which may be coloured or highlighted) and selectable foreground text. By the latter I mean foreground text that becomes highlighted as a result of user interaction (`TAB`-ing among the three panels or moving up/down in the appointments or todo list). The selectable foreground text "moves" attention from one part of the screen to another. To work well these three levels should be clearly distinguishable on the screen, in particular the highlighting should be clear. This may be trivial, but it is important that it works well.

Now to the separation of days. On my screen the highlighted day headings are prominent and draw the attention much more than any event separator. They are the only highlighted parts of the panel when it is not in focus. When the panel gets the focus (by `TAB`-ing) the currently selected item becomes highlighted and is easily spotted. Nothing new here. In particular, if the dayheading position is set "to the right", the selected item is the only highlighted part of the left part of the panel. I have no difficulty in picking out the agenda of each day, even when the empty lines between days are left out. But it would not be difficult to mark the day separation with a horisontal line.

And to the selected day. It has a much less prominant part to play now. Much more important is _the selected item_ (event/appointment) which indirectly determines the selected day. You operate on items (viewing, inserting, deleting, changing, moving them), not on days. You don't have to select a new day in order to operate on another item, you just scroll up or down in the appointments panel (unless it is far removed in time from the selected one in which case you do as previously and go to a new day). The calendar shows the context of the appointments panel, e.g. which day holds the selected item (and therefore is the selected day).

These are mostly personal impressions and opinions, and it would be most interesting to have other views expressed.

lhca:
> Another point is that the current day might be in the middle of the screen. I would probably find it easier to keep track of the currently selected day if it were always on top, even though that would mean less flexibility and remove the possibility the view all events of the past week.

That would go against the listbox concept, scrolling up and down a fixed list until you reach top or bottom. But all the movement commands (t, T, m, M, etc) and the arrow keys in the calendar bring the selected day to the top of the list.

src/day.c
lhca: > I found this confusing and reading the commit message or the comments did not help. What is the purpose of the new `order` argument? What does this double search do? Without reading the remaining code in detail: I guess the first loop tries to find an occurrence on the current day? But then, what does the other loop do? What exactly is the (new?) purpose of this function?

I'm not satisfied with this myself, and the comment is of the sort I try to avoid: saying what the code itself says.

The problem of setting the selected item correctly in the `day_items` vector (appointments list), when the vector/list is rebuilt, existed before. I changed the solution in commit https://github.com/lfos/calcurse/pull/193/commits/dd6a2a305091216a95b9cf65f303a929b0421d9b, but am no longer sure I like the way it was changed. I am going to look it over and update the PR as appropriate.

lhca:
I have reduced the number of commits further to ten and reshuffled them, hopefully making them more reviewable. The first four are preparatory and may be used in their own right without the last six, which contain the multiple days implementation.

The first two are concerned with the scrollwindow/listbox concepts. The first addresses an old issue in scrollwindows and is not required by the last six, the second is a follow-up for listboxes and _is_ required. See the commit messages.

The next two are minor corrections and _are_ required for the last six.

The multiple days feature is contained in the last six commits. There has been some refactoring and squashing, but the code is unchanged. There are new commit messages in several of them.

src/day.c
lhca: I rewrote it partly. See 4a19b95 and the commit message which has a comparison with "single day" calcurse and explains the differences.

lfos:
Thanks for continuing your work on this. Despite me not being very active on this PR lately, I would love to see this merged and I think we should make it the main focus of our next feature release.

I mentioned that I found the new UI experience to be not optimal. Playing around with it a bit more, I realized that one big thing that contributes to this is that it seems like the calendar panel can now be used to move in the calendar. But this is only true for a very limited range. Something I might be more comfortable with is a "sliding window", where the displayed days are changed once you reach the start/end of the list. We can discuss on what start/end means but something like not being able to go further down than the third-to-last day (same at the start of the list) might make sense. What do you think?

Also, we should not solely rely on colors to structure the output; calcurse has a monochrome mode where everything is black and white. It is probably not used very often but we should keep it as functional as we can if it does not cause any harm.

lhca:
> I mentioned that I found the new UI experience to be not optimal. Playing around with it a bit more, I realized that one big thing that contributes to this is that it seems like the calendar panel can now be used to move in the calendar.

[I assume that you mean "appointments panel", not "calendar panel".]

That is precisely the point of commit c8c3715. I quote the commit message: _This commit draws the full consequence: the selected day in the calendar is always (set to) the day of the selected item in the APP panel_.

> But this is only true for a very limited range.

???

> Something I might be more comfortable with is a "sliding window", where the displayed days are changed once you reach the start/end of the list.

??? But that's how it works already (or is supposed to work) as descibed in 55afa1b, "APP panel scrolling".

lfos:
I did not realize that! Maybe it is because scrolling only starts once you go "past" the last item. Can that be configured such that scrolling starts earlier? I also still don't like the visual separation, especially with the color scheme I would like to use. Maybe it would help to make it possible to use different colors for the heading and the current selection. But that can be added later. Will review the patches in detail now.

lfos:
One other thing is that the new list box implementation no longer behaves nicely when resizing windows. To demonstrate what I mean, start with an empty TODO list and add items 1 to 9 with item i having priority i (i.e. item 1 will be at the top of the list and item 9 at the bottom). Then select item 9. Resize the window such that only one item fits in the displyaed part of the list box. As before, the new implementation keeps the selected item in the displayed area. If you resize the window back to its original size (where all items are displayed), however, the list now keeps showing a single item only vs. restoring the full list in the previous implementation.

lhca:
> I did not realize that!  Maybe it is because scrolling only starts once you go "past" the last item.

Or maybe it's because you know too much ;-) An unsuspecting user would just press the down/up key.

> Can that be configured such that scrolling starts earlier?

Not easily.  It's a bit more complicated than it may look. If the range of days does not fit in the panel, the scrollling is just listbox scrolling. When you reach the last item in the list and press the down key, the move fails. This is detected, the range of days is adjusted and loaded, and the selection moved the beginning of the last day.

> I also still don't like the visual separation, especially with the color scheme I would like to use. Maybe it would help to make it possible to use different colors for the heading and the current selection. But that can be added later.

Notice that the separator between events and appointments has been shortened to leave a distance to the border. To make it less conspicuous it could be shortened further so that the beginning is aligned with the text above and below?

lhca:
> One other thing is that the new list box implementation no longer behaves nicely when resizing windows.

Interesting. Easily reproduced, also in the appointments panel (select the last item and resize twice).
The problem is in the scrollwindow when new contents are loaded or _when the window changes size_. In both cases the viewport (onto the pad) must be adjusted, in the first case because the number of pad lines changes (and the pad offset may change), in the second case because the pad offset changes (while the number of pad lines is unchanged).

I have fixed the bug by adding a call of `wins_scrollwin_set_pad()` to `wins_scrollwin_resize()`. This is the equivalent of the line `sw->line_num = sw->line_off = 0;` in `wins_scrollwin_init()` just above.

The fix has been incorporated in the first commit ("Scrollwindow pad improvement") of the series, and uploaded.

lfos:
Awesome, seems to work fine now! Some other things I noticed:

1. You changed the (meaning of) the colors in the calendar panel. Is there a specific reason for that? I know that the selected day now has a different marker. But why did you change the colors for the other indicators?
2. Having the week numbers next to the individual weeks is great but I think they are too close and easy to confuse with the numbers used for the days. I guess you will argue, again, that there are different colors but that is not enough distinction for my taste. I guess using (only) colors to structure output just doesn't work for me as well as it does for you.

lhca:
> You changed the (meaning of) the colors in the calendar panel. Is there a specific reason for that?

Oops. That was not on purpose. Part of a change in my personal version of calcurse crept in. I have added the missing part, which simplifies the colour code handling a little bit, and uploaded it (commit "Calendar display").

>  I know that the selected day now has a different marker.

The square brackets came about by chance. Originally I thought that the range of days in the appointments panel should be shown in the calendar, and used the brackets for that purpose. It turned out to be not very useful and rather distracting. The user's primary attention should be on the appointments panel and the selected _item_, not on the calendar panel and the selected _day_. And once you have settled on the number of days to load, the range of days is no longer of much importance, because the appointments panel works like an endless listbox in both directions. So I dropped the range from the calendar, but kept the brackets around the selected day. It has the advantage of not obscuring the day colour.

> Having the week numbers next to the individual weeks is great but I think they are too close and easy to confuse with the numbers used for the days.

That was also my first thought. Now I rather prefer the compact format, like a small table with row and column headings. Give it a try.

lfos:
Tried it for a while. Then I increased `weekw` from 2 to 3 which I like much better. Even if you say the leftmost column is a row heading, I think the columns (including the heading) should be evenly spaced.

Another thing that is irritating me is that days of the previous/next month are now included in the calendar view. There is really a lot of details being changed in this PR... I am not opposed to adding those days but would really prefer to have a visual distinction. Can we somehow "silence" all numbers not belonging to the current month? Gray instead of white?

lhca:
> Tried it for a while. Then I increased `weekw` from 2 to 3 which I like much better.

So do I. Distracted by other things, I never got back to it. I'll update the PR.

> I am not opposed to adding those days but would really prefer to have a visual distinction.

So would I, but I cannot think of a way that works.

lfos:
Additional comments:

* I think editing `general.multipledays` in the configuration menu should yield a prompt (just like `general.periodicsave`). Having to press the edit key 20 times to decrease the value is super annoying.
* Note that when setting `general.multipledays` to `1` to restore the original behavior (no multi-day display), empty days are now marked with the `(none)` marker. I don't think this is necessarily a bad thing; it is at least consistent. But it might confuse users unaware of the multi-day setting. We probably want to make multi-day the default in the future, though, so it should be all fine.
* Some of the commits need better subject lines. You should be able to tell (very roughly) what a commit does by just reading that single line. At least for the `Calendar display` patch, this is certainly not possible. Optimally, you should also try to stick to our commit message guidelines [1], use imperative tense of the subject line and proper formatting. I won't be too strict about that, it is just a general note!

[1] https://github.com/lfos/calcurse/blob/pu/doc/submitting-patches.txt#L103

src/utils.c
lfos: Yes, `ui_get_slctd_day()` sounds good.

src/calcurse.c
```diff
@@ -143,7 +143,7 @@ static inline void key_add_item(void)
 
 static inline void key_edit_item(void)
 {
-	if (wins_slctd() == APP) {
+	if (wins_slctd() == APP && !event_dummy()) {
```
lfos: This does not look right. And it does not work. This commits allows you to edit the description of a dummy item (which makes calcurse crash). You probably meant `ui_day_dummy()` here?

src/ui-day.c
```diff
@@ -39,44 +39,35 @@
...
+void ui_day_find_sel(void)
 {
-	return update_time_in_date(ui_day_selitem()->order, 0, 0);
-}
-
-/*
- * Return true if the selected APP item is the dummy event.
- */
-int ui_day_dummy(void)
-{
```
lfos: Huh? Why are you introducing `ui_day_dummy()` only to remove it in a follow-up commit? I guess we want to keep this function and use it instead of the faulty event_dummy() calls I commented on.

src/calcurse.c
```diff
@@ -154,7 +154,7 @@ static inline void key_add_item(void)
 
 static inline void key_edit_item(void)
 {
-	if (wins_slctd() == APP && !event_dummy()) {
+	if (wins_slctd() == APP && !event_dummy(ui_day_get_sel())) {
```
lfos: Huh? This looks like a bug fix to me that should not belong in this commit. It should be part of the commit that introduced the faulty code. In fact, I would prefer keeping `ui_day_dummy()` and using it already in the first commit that introduced `event_dummy()` here. This should allow us to get rid of all the stray bug fixes.

src/ui-day.c
```diff
@@ -39,44 +39,35 @@
...
+	int n;
 
-	return d ? d->start == DUMMY : 0;
-}
-
-void ui_day_set_selitem_by_aptev_ptr(union aptev_ptr p)
-{
-	int n = day_get_position_by_aptev_ptr(p);
-	if (n >= 0)
+	if ((n = day_sel_index()) != -1);
```
lfos: This looks suspicious. Are you sure you want that semicolon?

lhca:
> I also still don't like the visual separation, especially with the color scheme I would like to use. Maybe it would help to make it possible to use different colors for the heading and the current selection.

Another possibility would be to make the DAY_SEPARATOR visible. At present it is invisible with a configurable height of 0, 1 or 2 lines. It could be drawn horisontally from border to border with any character (ACS_HLINE, ACS_BULLET, '=', ...) and possibly highlighted?

Or the EVENT_SEPARATOR could be made less conspicuous by replacing the character ACS_HLINE with `'-'` or ACS_BULLET?

lfos:
> Another possibility would be to make the DAY_SEPARATOR visible. At present it is invisible with a configurable height of 0, 1 or 2 lines. It could be drawn horisontally from border to border with any character (ACS_HLINE, ACS_BULLET, '=', ...) and possibly highlighted?
> 
> Or the EVENT_SEPARATOR could be made less conspicuous by replacing the character ACS_HLINE with `'-'` or ACS_BULLET?

I would try using `ACS_HLINE` to separate days and an empty line to separate events from appintments within the same day.

src/calcurse.c
lhca: Why I you reviewing commits that were long since replaced? This and your two previous comments concern outdated sources. Please take the commits of 17 March as the starting point.

lfos: Most likely because I am not using GitHub review in the way I should. That's what GitHub gave me when clicking on the commit summary of this PR.

The very same issues are still there in the most recent version of the PR, aren't they?

src/ui-day.c
lhca: A genuine bug. Corrected in next upload.

src/calcurse.c
lhca: No!

lhca:
> I would try using `ACS_HLINE` to separate days and an empty line to separate events from appintments within the same day.

I have uploaded a correction to that effect and until further kept it in a separate commit.

src/calcurse.c
lhca: See below.

src/ui-day.c
lhca: See below.

src/calcurse.c
lhca: _In answer to this and the two previous comments_:

The function `ui_day_dummy()` was introduced from the start. Its purpose was to tell whether the selected item (in the appointments panel) is the dummy event; it had no arguments. Later it got a single one, as a need turned up for testing any item, not just the selected one. This argument is needed in my private version only (where development takes place), but can be left out of the GitHub version. Still later the function was renamed to `event_dummy(struct day_item *)` and moved from `ui-day.c` to `event.c` where the external variable `struct event dummy` is declared.

During the process of rebasing and squashing commits before upload, conflicts of course turned up that had to be fixed. When commits are reordered in interactive rebasing, function name changes are a particularly nasty business. I can no longer recall what happened or what I did, but `ui_day_dummy()` should have disappeared completely. In stead it survived in commit 7cba46d without being called and was removed in the following cf86e9e; `event_dummy()` was defined as `int event_dummy(struct day_item *item) {...}`, but called without argument; the forward declaration in `calcurse.h` was `int event_dummy();`. The code compiles, and I must have assumed that everything was OK. The same state remained in the following commit cf86e9e, before being corrected in 7c6488d.

I propose the following:

1. `ui_day_dummy()` is removed.
2. The `event_dummy()` function keeps its argument (it may come in handy), and the declaration in `calcurse.h` is corrected.
3. The faulty calls of `event_dummy()` are corrected.

lfos: I agree with your analysis and your suggestion. Keeping the version with an argument seems to make sense, particularly when you say there are other patches needing this (that you may or may not want to merge later).

lfos:
Awesome! I really like the new UI epxerience, especially with `appearance.dayseparator = 1`. I think this is fairly close to being ready for `pu`.

Two minor things:

* It is pretty counterintuitive that the separator magically disappears *and* the former separator magically appears with `appearance.dayseparator = 0`, though. Suggestions: Drop the option to set `appearance.dayseparator = 2` (does not make sense to allow spacing of 1 or 2 lines if we do not make it fully configurable). Then make `appearance.dayseparator` a Boolean option that switches between day separator and event/appointment separator.
* I don't think we should have a day separator after the last day. We also hide the event/appointment separator if one of the lists is empty.

And two remarks from earlier still apply:

* The `appearance.dayseparator` should be configurable with a prompt, just like `appearance.dayseparator`. Having to press the edit key 20 times just to decrease the value by 1 is pretty annoying.
* The subject lines of some of the commit messages are not very meaningful. `Calendar display` does not tell us much. Maybe at least go with `Refactor calendar display code`. I can also change the commit message when merging this series if you prefer that!

lhca:
> Awesome! I really like the new UI epxerience, especially with `appearance.dayseparator = 1`. I think this is fairly close to being ready for `pu`.
>

The patch wasn't meant to be the final implementation, just an easy way of getting your opinion. I'm glad you like it.

So the visible day separator is definitely in. Should it be highlighted as the day heading? Probably not.
 
> Two minor things:
> 
>  It is pretty counterintuitive that the separator magically disappears _and_ the former separator magically appears with `appearance.dayseparator = 0`.

??? appearance.dayseparator is the number of lines of the separator (the height). Only the last of those have a visible, horisontal line drawn. With height set to zero there can be no visible (or empty) line. So how could it appear again? The empty line you see above the day heading is probably the empty line after the last entry of the day (either event or appointment). Do you have `appearance.emptyline = yes`? The background for that is in the commit message ee89de0.

I never felt the need for a visible day separator myself and made it up to the user to choose between 0, 1 or 2 (empty) lines. I think all combinations should be possible: visible or invisible day separator of heigth 0, 1 or 2, except that you cannot have a visible separator of zero height. That gives five possibilities, and `appearance.dayseparator` could be made to cycle through them (an easy and foolproof method that I like). Or it could be two configuration variables, but that will be more complicated because of the dependence mentioned.

At present, the event separator always has height 1, when inserted, but that could also be dropped to allow people who like a crammed panel to load even more. Then it needs its own independent `appearance.eventseparator` variable that could cycle through `visible/empty/none`.

>  I don't think we should have a day separator after the last day. We also hide the event/appointment separator if one of the lists is empty.

Conceptually there is no last day. The illusion is that of an endless list. You mean, of course, the last day of the panel. But that may not be the last day of the listbox which may have more days in the scroll window. Would it be worth the trouble to try to detect whether the last line of the view port of the scroll window contains a visible day separator line and then scroll it out of sight? That would bring in a line or lines at the top, and who knows what is then at the bottom. I don't think it's possible.

lfos:
> ??? appearance.dayseparator is the number of lines of the separator (the height). Only the last of those have a visible, horisontal line drawn. With height set to zero there can be no visible (or empty) line. So how could it appear again? The empty line you see above the day heading is probably the empty line after the last entry of the day (either event or appointment). Do you have `appearance.emptyline = yes`? The background for that is in the commit message [ee89de0](https://github.com/lfos/calcurse/commit/ee89de0243c2af374ad4a89e4c6ce301f2e67af2).

Sorry, I was a bit vague there. If `appearance.dayseparator` is strictly positive, there is a separator between days but no separator betweens events and appointments of the same day (which is good). If `appearance.dayseparator` is zero, there is no separator between days (which is also good) but suddenly a separator betweens events and appointments of the same day appears which is not very intuitive.

> I never felt the need for a visible day separator myself and made it up to the user to choose between 0, 1 or 2 (empty) lines. I think all combinations should be possible: visible or invisible day separator of heigth 0, 1 or 2, except that you cannot have a visible separator of zero height. That gives five possibilities, and `appearance.dayseparator` could be made to cycle through them (an easy and foolproof method that I like). Or it could be two configuration variables, but that will be more complicated because of the dependence mentioned.

Again, I don't see why we should allow separators of height 0, 1 and 2 but not of height 3. I also don't think you need any additional space if you already have the separator line but that may be subjective.

Your suggestion to add two variables also makes sense to me. One Boolean variable can be used to decide whether the separator line appears between days or between events/appointments. Another variable that is configurable via a numerical prompt to set the number of empty lines between days. I would not count the separator line as an empty line (since it is clearly not empty) which also solves the dependence issue you mentioned.

Alternatively, use a single variable for the day separator and toggle between `none`, `line`,  `1 empty` and `2 empty`. Would not allow for using more than two empty lines but maybe it's worth the trade-off.

> At present, the event separator always has height 1, when inserted, but that could also be dropped to allow people who like a crammed panel to load even more. Then it needs its own independent `appearance.eventseparator` variable that could cycle through `visible/empty/none`.

This would work well with the `none`, `line`,  `1 empty` and `2 empty` suggestion above. We would have the same options for the event/appointment separator and the day separator.

> 
> > I don't think we should have a day separator after the last day. We also hide the event/appointment separator if one of the lists is empty.
> 
> Conceptually there is no last day. The illusion is that of an endless list. You mean, of course, the last day of the panel. But that may not be the last day of the listbox which may have more days in the scroll window. Would it be worth the trouble to try to detect whether the last line of the view port of the scroll window contains a visible day separator line and then scroll it out of sight? That would bring in a line or lines at the top, and who knows what is then at the bottom. I don't think it's possible.

I actually mean the last day of the list box, not the last day in the visible part of the list box. That should be easy to detect.

lhca:
I fear there is case of miscommunication and terminolgy mixup. I'll try again and use my words even more carefully. Remember that at this point I only wanted to hear your opinion of the horisontal line at the end of the day and the empty line between events and appointments. There accidental appearance or disappearance in other cases is of no significance.

As to terminlogy. When I say "separator", I mean one or several lines that separates lines above from lines below. A line of a separator may be either empty (blank) or have visible (non-blank) characters drawn on it, for example a horisontal line from border to border. By a visible separator I mean a separator with at least one line with visible characters.

> > ??? appearance.dayseparator is the number of lines of the separator (the height). Only the last of those have a visible, horisontal line drawn. With height set to zero there can be no visible (or empty) line. So how could it appear again? The empty line you see above the day heading is probably the empty line after the last entry of the day (either event or appointment). Do you have `appearance.emptyline = yes`? The background for that is in the commit message [ee89de0](https://github.com/lfos/calcurse/commit/ee89de0243c2af374ad4a89e4c6ce301f2e67af2).
> 

> Sorry, I was a bit vague there. If `appearance.dayseparator` is strictly positive, there is a separator between days but no separator betweens events and appointments of the same day (which is good).

You're still vague. There is an _event_ separator between events and appointments: the empty line.

> If `appearance.dayseparator` is zero, there is no separator between days (which is also good) but suddenly a separator betweens events and appointments of the same day appears which is not very intuitive.

The event separator has been there all the time, but is now drawn with a horisontal line. But that's an accidental fact that you must disregard. The intention is to make such things configurable. 
> 
> > I never felt the need for a visible day separator myself and made it up to the user to choose between 0, 1 or 2 (empty) lines. I think all combinations should be possible: visible or invisible day separator of heigth 0, 1 or 2, except that you cannot have a visible separator of zero height. That gives five possibilities, and `appearance.dayseparator` could be made to cycle through them (an easy and foolproof method that I like). Or it could be two configuration variables, but that will be more complicated because of the dependence mentioned.

> Again, I don't see why we should allow separators of height 0, 1 and 2 but not of height 3.

Not allowing 2 with the argument "then why not 3" is beyond me. Or in other words, either 1 or infinity?

> I also don't think you need any additional space if you already have the separator line but that may be subjective.

Exactly. _Others_ may need additional space, and _I_ think that 2 is reasonable leeway.
 
> Your suggestion to add two variables also makes sense to me. One Boolean variable can be used to decide whether the separator line appears between days or between events/appointments. Another variable that is configurable via a numerical prompt to set the number of empty lines between days. I would not count the separator line as an empty line (since it is clearly not empty) which also solves the dependence issue you mentioned.

You're much too vague again for my taste. If by "separator line" you mean a horisontal line, you are saying that I should live with the damn thing always? No way!

> Alternatively, use a single variable for the day separator and toggle between `none`, `line`, `1 empty` and `2 empty`. Would not allow for using more than two empty lines but maybe it's worth the trade-off.

That's my own suggestion, almost. The only missing case is `empty+line`, but I can do without that. No one has asked for more that two empty lines.
 
> > At present, the event separator always has height 1, when inserted, but that could also be dropped to allow people who like a crammed panel to load even more. Then it needs its own independent `appearance.eventseparator` variable that could cycle through `visible/empty/none`.
>
> This would work well with the `none`, `line`, `1 empty` and `2 empty` suggestion above. We would have the same options for the event/appointment separator and the day separator.

`line` may be a better choice than `visible`, but `2 empty` is a surprise.

> > > I don't think we should have a day separator after the last day. We also hide the event/appointment separator if one of the lists is empty.
> > 
> I actually mean the last day of the list box, not the last day in the visible part of the list box. That should be easy to detect.

But why? A single line at the bottom of the panel may either be in the middle of the list or at the end of the last day of of the list, and you won't know which (unless you look very carefully at the scrollbar). And you shouldn't care - the list is conceptually endless.

lfos:
> As to terminlogy. When I say "separator", I mean one or several lines that separates lines above from lines below. A line of a separator may be either empty (blank) or have visible (non-blank) characters drawn on it, for example a horisontal line from border to border. By a visible separator I mean a separator with at least one line with visible characters.

That's indeed not the terminlogy I used but we can agree on that!

> The event separator has been there all the time, but is now drawn with a horisontal line. But that's an accidental fact that you must disregard. The intention is to make such things configurable.

That's exactly what I meant (using my terminlogy). It's great to hear you agree that the current behavior should change.

> > Again, I don't see why we should allow separators of height 0, 1 and 2 but not of height 3.
> 
> Not allowing 2 with the argument "then why not 3" is beyond me. Or in other words, either 1 or infinity?

What I suggested was allowing arbitrary integers that can be entered in a prompt. I did not mean to suggest dropping `2`. I just did not understand why we should limit ourselves to `1` and `2` when the very same implementation would have allowed for arbitrary positive numbers with now extra maintenance cost.

> Exactly. _Others_ may need additional space, and _I_ think that 2 is reasonable leeway.

Yes, there is always an arbitrary choice if you do not allow arbitrary integers. But we can keep that for now. Let's stop bikeshedding.

> 
> > Your suggestion to add two variables also makes sense to me. One Boolean variable can be used to decide whether the separator line appears between days or between events/appointments. Another variable that is configurable via a numerical prompt to set the number of empty lines between days. I would not count the separator line as an empty line (since it is clearly not empty) which also solves the dependence issue you mentioned.
> 
> You're much too vague again for my taste. If by "separator line" you mean a horisontal line, you are saying that I should live with the damn thing always? No way!

Huh? I am slightly confused. What I suggested is having one configuration variable to enable/disable the visible horizontal line and another configuration variable to add a configurable number of additional empty lines to the separator. So the visible line is either turned off and there is only `n` empty lines between days (where `n` is configurable) or the visible line is turned on and then, there is the visible line plus `n` empty lines. Maybe my comment makes more sense with this additional explanation.

>> Alternatively, use a single variable for the day separator and toggle between `none`, `line`, `1 empty` and `2 empty`. Would not allow for using more than two empty lines but maybe it's worth the trade-off.
> 
> That's my own suggestion, almost. The only missing case is `empty+line`, but I can do without that. No one has asked for more that two empty lines.

Great! As I mentioned before, I am fine with that implementation as well if you prefer that!

> > This would work well with the `none`, `line`, `1 empty` and `2 empty` suggestion above. We would have the same options for the event/appointment separator and the day separator.
> 
> `line` may be a better choice than `visible`, but `2 empty` is a surprise.

What do you mean? `2 empty` refers to "two empty lines" as a separator. `2 empty` was just a suggestion without thinking about naming too much. If you think it is not intuitive, feel free pick anything else! Maybe `2 empty lines`?

> 
> > > > I don't think we should have a day separator after the last day. We also hide the event/appointment separator if one of the lists is empty.
> > 
> > 
> > I actually mean the last day of the list box, not the last day in the visible part of the list box. That should be easy to detect.
> 
> But why? A single line at the bottom of the panel may either be in the middle of the list or at the end of the last day of of the list, and you won't know which (unless you look very carefully at the scrollbar). And you shouldn't care - the list is conceptually endless.

Because the visible horizontal lines are visual separators of visible content. They are part of the UI and not part of the underlying logical structure. Just like I am not putting two extra newlines at the end of this comment because I know you will reply and this is not the end of the conversation. I only add empty lines to make this comment more readable. And if the comment were in a scrollable text box, I would still add empty lines between paragraphs only. But again, I don't feel strongly about it, so if you insist on having the line at the end, keep it!

lhca:
Back to the "real thing". I have had a new idea that I hope will make the day, if you know what I mean, and may put an end to a not very interesting exchange.

Instead of putting the horisontal line at the end of the day in the day separator (which lead to minor complications), put it at the start of the day in the day heading (which is simple) and leave out the visible event separator. The advantages are

- it gives the visual distinction you are looking for (I hope)
- it looks nice
- it saves a line per day when `appearance.dayseparator` is 0
- the three elements DAY_HEADING, EVNT_SEPARATOR and DAY_SEPARATOR are independent of each other and may be configured separately
- every day has a day heading

I have uploaded a mockup that cannot be configured. What do you think?

lfos:
Looks good to me. Let's give it a try!

lhca:
Ok. Here are some further thoughts.

The english word "line" has two different meanings (among others): a line of text (in german "eine Zeile") or a geometric line (in german "eine Linie" or "eine Gerade"). I will write either "text line" or "line segment" as needed to avoid ambiguities. I assume that events are presented first, followed by appointments (as before).

There are then four elements that influence the visual presentation of each day:

- header
- separation of events from appointments
- separation of appointments from each other
- end of day (footer)

The header contains by default the day heading only (the day's date), but a line segment may be displayed to emphasize the beginning of each day. The date and the line segment may share a single (text) line, or the line segment may have a line of it's own below the date line [_should both be available as configuration alternatives? Or only one of them? If so which one? I'm inclined towards the line segment below the date_]. The default configuration is day heading with no line segment [_or should the default be to show a line segment?_].

Events are by default separated from appointments by an empty line, but it may be left out. A line segment as separation is no longer possible because it makes the separation of the panel into days less clear.

An appointment takes up two text lines by default followed by an empty line that may be left out.

The end of a day is by default an empty line that may be left out.

The last three elements (event separator, empty appointment line and end of day separator) are each configured with a boolean. The header is either configured with a boolean (line segment/no line segment) or with date only/date and line/date above line.

lfos:
Everything you suggest sounds good to me. If the put the date line and the line segment in separate lines, I would put the date *after* the line segment, though (it separates days and the date line belongs to the next day).

I would also show the line segment by default: all other options that affect UI currently default to showing extra information and visual spearations where possible (the separator between events and appointments, compact mode is disabled by default, system dialogs are enabled). That being said, I don't use these settings myself and would be open to switching the default behavior. But let's do that in a separate PR and try to get this merged. There is too much going on here already.

lhca:
Uploaded.

> If the put the date line and the line segment in separate lines, I would put the date _after_ the line segment, though (it separates days and the date line belongs to the next day).

I have done as you suggested. You might argue that the line segment _emphasizes_ the day heading, and I think it looks less pleasing with two parallel lines at the top, but see for yourself. I don't care either way.

 > I would also show the line segment by default: all other options that affect UI currently default to showing extra information and visual spearations where possible (the separator between events and appointments, compact mode is disabled by default, system dialogs are enabled).

Done as suggested.

I plan on squashing the two newest commits into the earlier ones, while also attending to your still outstanding comments.

lfos:
Great! I would simply add an exception and drop the line segment for the first day. Yes, it is a bit inconsistent but that line is not helping with anything.

Thanks for all your work!

src/utils.c
lhca: I would rather not introduce name changes at this stage (experience has taught me that). Both variables should be looked at when this PR is closed.

src/ui-day.c
lhca: Fixed.

lhca:
> And two remarks from earlier still apply:
> 
> The `appearance.dayseparator` should be configurable with a prompt, just like `appearance.dayseparator`. Having to press the edit key 20 times just to decrease the value by 1 is pretty annoying.

Yes, it is a bit annoying, but only a little bit. And you are not likely to do it very often. On the positive side, it is very simple and foolproof, also for the user. You just keep pressing the same key, no additional prompt, error check or error message.

On the other hand,  `general.periodicsave` (which I assume is what you meant) uses `atoi()` in reading an integer. This function has a particularly bad interface and does not report any errors. If you type anything but an integer, `atoi()` returns the valid value zero, and periodic save may inadvertently be turned off. Here zero is not a valid value, so `atoi()` is not quite as useless. But I admit that 21 is testing your patience and cannot be raised by much.

That brings up a much more important question: should there be an upper limit? And what should it be? We have discussed it previously, and you were against setting any limit on the number of days. Is that still the case, now that you have tried it?

> The subject lines of some of the commit messages are not very meaningful. `Calendar display` does not tell us much.

True. For "Calendar display" I propose "Add week numbers for all weeks, full weeks for first and last". I can easily turn most of the subject lines into the imperative tense.

lfos:
> On the other hand, `general.periodicsave` (which I assume is what you meant) uses `atoi()` in reading an integer. This function has a particularly bad interface and does not report any errors. If you type anything but an integer, `atoi()` returns the valid value zero, and periodic save may inadvertently be turned off. Here zero is not a valid value, so `atoi()` is not quite as useless. But I admit that 21 is testing your patience and cannot be raised by much.

I think I could live with the current implementation if there at least were a way to decrease the value. The first time I wanted to switch the value from 7 to 1, I went past 1 to 2. That means I had to press the edit key 36 times in total to set the value which is insane.

That being said, if you think changing this at the current stage of the PR is messy, we can also keep it and unify *all* integer prompts in a follow-up patch.

> That brings up a much more important question: should there be an upper limit? And what should it be? We have discussed it previously, and you were against setting any limit on the number of days. Is that still the case, now that you have tried it?

I still prefer not having an upper limit (i.e. a rather big upper limit since we cannot store arbitrarily large integers anyway). As I made clear before, I am saying this not because I dislike the current implementation. My reasoning is based on the absence of an explanation why we should have such a limit in the first place.

Allowing 22 days is no harder to implement than allowing 21 days. And allowing 30 days is no harder than allowing 22 days. And I could potentially see somebody with a huge display wanting to display 30 days instead of 21. I don't see why the user's freedom of choice should be limited just because we feel like 21 days sounds enough for us. I you insist on a limit, 365 days might be a sensible choice, and I can't see anybody realistically wanting to go past that in the near future.

lhca:
> I still prefer not having an upper limit (i.e. a rather big upper limit since we cannot store arbitrarily large integers anyway). As I made clear before, I am saying this not because I dislike the current implementation. My reasoning is based on the absence of an explanation why we should have such a limit in the first place.

As to an explanation, I gave that in an earlier exchange (https://github.com/lfos/calcurse/issues/152#issuecomment-453961893).

Your reasoning is a very general and my general answer is: you should test what you implement. If you allow the user to input 365, you should know that it works. But back to the specific case, calcurse.

Nothing is gained by loading days that are not displayed. Calcurse will load new days as needed. Taking this fact to its logical conclusion, I have realised that `general.multipledays` should be removed, not limited.

There is no reason to ask the user how many days to load, it's none of his business, and he shouldn't be bothered with the question. Think about it. The starting point was to show more than one day or, otherwise put, fill the appointments panel with days. Calcurse can do that from information readily available without asking the user: 

- the number of lines in the panel from the screen size, `appearance.notifybar` and `appearance.compactpanels`
- the minimum number of lines per day from `appearance.headerline`, `appearance.eventseparator`, `appearance.dayseparator` and `appearance.emptyline`.

Then a simple division tells how many days to load. The number will always be an overestimation.

This is very easy to implement. The function `day_get_days()` just returns the result of the calculation instead of `general.multipledays`. Different display sizes are automatically taken into account, including terminal resizing. The method can be refined to take account of the number of items on the agenda of each day.

lfos:
> As to an explanation, I gave that in an earlier exchange ([#152 (comment)](https://github.com/lfos/calcurse/issues/152#issuecomment-453961893)).
>
> Your reasoning is a very general and my general answer is: you should test what you implement. If you allow the user to input 365, you should know that it works. But back to the specific case, calcurse.

I still don't see how that adds anything to the discussion. An argument for a limit describes a situation where the limit is strictly better than no limit. In the linked comment, you are merely saying that most of the time, having no limit is not better than having a limit. In the second part you are saying that we need to test our software. We have to test a limit of 365 just like a limit of 21.

I think this discussion is slowing down way things much more than it should. My suggestion for you is to reconsider what I said again from a neutral perspective and then make a decision. I will accept that decision and merge the PR. It's not like either choice would hurt the project. In the worst case, users will dislike the implementation and create a bug report. None of the things we are discussing are part of a central design decision that cannot be reverted later.

> Nothing is gained by loading days that are not displayed. Calcurse will load new days as needed. Taking this fact to its logical conclusion, I have realised that `general.multipledays` should be removed, not limited.
> 
> There is no reason to ask the user how many days to load, it's none of his business, and he shouldn't be bothered with the question. Think about it. The starting point was to show more than one day or, otherwise put, fill the appointments panel with days. Calcurse can do that from information readily available without asking the user:
> 
> * the number of lines in the panel from the screen size, `appearance.notifybar` and `appearance.compactpanels`
> * the minimum number of lines per day from `appearance.headerline`, `appearance.eventseparator`, `appearance.dayseparator` and `appearance.emptyline`.
> 
> Then a simple division tells how many days to load. The number will always be an overestimation.
> 
> This is very easy to implement. The function `day_get_days()` just returns the result of the calculation instead of `general.multipledays`. Different display sizes are automatically taken into account, including terminal resizing. The method can be refined to take account of the number of items on the agenda of each day.

Sounds like a good idea to me. The only downside is that users will no longer be able to disable the "multiple days" feature, right? Some users might prefer a very clear UI that only shows items of one day.

lhca:
> Sounds like a good idea to me. The only downside is that users will no longer be able to disable the "multiple days" feature, right? Some users might prefer a very clear UI that only shows items of one day.

Let `general.multipledays` stay as a boolean that turns the feature on/off.

lhca:
Uploaded what is hopefully ready for merging into pu.

The subject lines have been amended, the `general.multipledays` is turned into a boolean that enables the feature, the number of days to load is calculated automatically, and the horisontal line that introduces a day is not drawn for the first day loaded (and therefore not present when multiple days are turned off).

The five configuration variables that control the feature are all booleans, and all are true by default.

lfos:
Awesome! There are some minor things I would like to improve but all that can be done on top of this patch set. Merged into pu. Thank you again for investing so much time in implementing and reworking this!

PR #195, opened 2019-03-07
calcurse-vdirsyncer improvements

vxid:
I'm opening this PR following the discussion in #35 to change the current behavior of calcurse-vdirsyncer, most notably by disabling any direct vdirsyncer interaction at all.

This is what I had in mind:

The script is either called with an `import` or `export` command, and only executes same command.
Old behaviour could be obtained calling something like:

`calcurse-vdirsyncer export <vdir> && vdirsyncer sync && calcurse-vdirsyncer import <vdir>`


Tonus1:
That sounds good to me. I was just wondering if import and export might create duplicate entries.

vxid:
I tested the new version and no, it does not create duplicate entries since uids are checked. The script also doesn't delete any data unless asked to.

vxid:
I pushed the new version of the script.

@lfos @Tonus1  I agree that it is much more flexible now. Do you have any other improvement ideas?

lfos:
Looks good to me, I merged the patch into pu. Should we still call the script `calcurse-vdirsyncer`, though?

vxid:
Thanks.
I think we can keep `calcurse-vdirsyncer` or use `calcurse-vdir` since the scrips is designed to work
with the vdir format.

lfos:
Suggestion: Rename both the script and the directory to `calcurse-vdir`. Add a `calcurse-vdirsyncer` shell script that does what the first version of the script did: run `calcurse-vdir` twice and `vdirsyncer` inbetween.

I guess I will squash the commits into a single commit adding `calcurse-vdir` support before merging into master.

lfos:
We should mention both `calcurse-vdir` and `calcurse-vdirsyncer` in the man page, of course.

vxid:
I will rename the script to calcurse-vdir.

As for adding a different script with a different name (which might be confusing), why not simply mention how to chain operations it in the README? I am using a simple alias that works really well:
```sh
alias calsync='calcurse-vdirsyncer export $VDIR_HOME -f -D $CALCURSE_HOME && \
	vdirsyncer sync && \
	calcurse-vdirsyncer import $VDIR_HOME -f -D $CALCURSE_HOME'
```

lfos:
> As for adding a different script with a different name (which might be confusing), why not simply mention how to chain operations it in the README? I am using a simple alias that works really well:

I guess it's a matter of taste. One reason for adding a script is that it works out of the box and that a common use case is running the synchronization script as a cron job where a wrapper script comes in handy.

vxid:
Fair enough. It may take me a while to write it though, mainly because I want to make sure that nothing breaks when syncing while calcurse is running and the necessary checks to the script. Do you think that could happen in another PR?

lfos:
It could but it can also happen here. I don't think you need to take special care of that. Just put the three lines into a script (with proper command line arguments and quoting).

vxid:
I created the script and tested it.
I also added it to the Makefile.

lfos:
The script belongs to `contrib/vdir/`; `scripts/` is not for contrib stuff.

lfos:
There also is a whitespace error in the usage line (spaces instead of tabs), and you should quote at least `$VDIR` and `$DATADIR`.

vxid:
My bad. I changed the script's location and fixed the quoting issues.

lfos:
I made some minor modifications and squashed all your patches into a single commit adding the scripts. Merged that into pu. Thanks for all your work!

Issue #196, opened 2019-03-08
Screen redrawing issues in calendar weekly view


vxid:
I noticed that the calendar pane is not immediately redrawn when deleting an event, leading to some deleted events still appearing in the weekly view and other artifacts.

In the following screenshot, I deleted the only event I had on March 14th. It still shows up in the weekly view and, for some reason, another event on Sunday 17th is highlighted as well:
![2019-03-08-115433_1920x1080_scrot](https://user-images.githubusercontent.com/16440823/54024946-601fa880-4199-11e9-8359-b3d443a98d24.png)

Bonus bug: when switching between weekly and monthly view from the calcurse interactive config, redrawing seems to completely give up.

![2019-03-08-115324_1920x1080_scrot](https://user-images.githubusercontent.com/16440823/54025101-cad0e400-4199-11e9-8fe4-12b2f9723fcb.png)

lhca:
I believe the two cases are unrelated and will open a PR that fixes the first one.

The second one is strange. Does it also happen if you switch views with Next View/Previous View (`^N`/`^P`). Do you mind disclosing something about your setup? OS, terminal emulator, calcurse version?

vxid:
@lhca The second one only seems to happen when switching from weekly to monthly.
I use Arch Linux, urxvt and built calcurse from the latest master.

lhca:
The two cases _were_ related after all, see updated #197 which fixes both.

This goes to prove that weekly view isn't much used.

PR #197, opened 2019-03-08
Fix calendar weekly view

lhca:
Addresses #196.


lfos:
Merged into pu, thanks!

PR #198, opened 2019-03-11
Fix compile issues in doc directory

bmunger:
This solves issue with the Makefile in the doc directory.  Some files have been renamed or converted and not reflected in the m4 file.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -92,6 +92,11 @@ if test x"$use_asciidoc" = x"auto"; then
   if test -z "$ASCIIDOC"; then
     have_asciidoc=no
     AC_MSG_WARN([AsciiDoc not found - cannot rebuild documentation!])
+    if test -f "doc/manual.html" && test -f "doc/submitting-patches.html"; then
+      AC_MSG_WARN([Found pre-built documentation.])
+    else
+      AC_MSG_ERROR(Asciidoc is required to build documentation.)
+    fi
   else
     have_asciidoc=yes
   fi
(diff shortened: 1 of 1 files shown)


lfos:
Thank you for opening this PR. I think you are addressing a long-standing issue.

Unfortunately, it is not that simple, though: currently, the release tarballs include pre-built documentation and *should* allow for performing a full build (including documentation) without having AsciiDoc installed. If it does not, please let me know. When using a development snapshot, AsciiDoc is a requirement for building documentation.

Therefore, the build system currently checks for AsciiDoc but only warns when it is not available (since a build running from a release tarball will complete successfully anyways). I see two options:

1. Add an additional check to the `configure` and error out if neither AsciiDoc nor pre-built documentation is available.
2. Make AsciiDoc a requirement, even for release tarballs.

I prefer the first option. Opinions?

vxid:
I think the first option is preferable as well.

bmunger:
Option 1 should be fine.  configure should not complete without error in a condition where make will fail.

lfos:
Are you interested in replacing this PR with a patch that implements this?

bmunger:
Sure no problem.  Soon as I get home from work I will revert and add the conditional checks to configure.ac.  Thanks.

bmunger:
Reverted previous change.  This change to configure.ac will warn when asciidoc and a2x have not been found but will continue if pre-built documentation and man pages are found, if not it will produce an error.  I don't expect that the a2x error will ever be seen since it's part of the asciidoc package though.  I opted to use shell functions to test for the pre-built files instead of AC_CHECK_FILE due to some concerns over cross compile during package building which may happen in some Linux distributions.  The changes have been tested without asciidoc and without the pre-built files, without asciidoc and with pre-built files, and with asciidoc.  It behaves as expected.  Let me know if this is acceptable.  Thanks.

lfos:
Merged into pu with some minor changes. Thanks!

PR #199, opened 2019-03-13
Change calcurse -C and -D long options

vxid:
See #192


lfos:
Merged into pu, thanks!

Issue #201, opened 2019-03-17
Deadlock when leaving configuration menu


lhca:
The recently released version 4.4.0 has exposed a deadlock that unfortunately is present already in version 4.3.0.

### Problem
Before 4.4.0, on startup calcurse would issue a fatal error and exit, if the configuration file `conf` contained an invalid entry. The same would happen when calcurse saved the configuration file, except it never got that far. That could only happen if you edited the file and introduced an error after calcurse had loaded it - a very unlikely scenario. But in that case calcurse would deadlock on exit for reasons explained below.

In 4.4.0, the fatal error on startup is relaxed to a warning and calcurse ignores the variable and  continues thus exposing the deadlock. If you in that situation enters the configuration menu, calcurse will deadlock on exit from the menu when it tries to save the configuration (which happens on menu exit and no longer on program exit or when the save command is issued).

### Reason
The deadlock is caused by the main thread waiting for the notification thread (which keeps the clock ticking) to cancel. That thread is blocked (notice the stopped clock) trying to lock a mutex held by the main thread.

### Workaround
If calcurse complains about a configuration option on startup, exit, open the configuration file in an editor and remove the offending option or comment it out with a `#`.

### Proposed solution
When saving a defective configuration file, calcurse should warn (as on startup) and comment the option out in the saved file.

Additionally the use of the mutex (called `nbar.mutex`) should be reviewed. I believe it is superfluous when loading/saving the configuration file.

lfos:
Merged into pu, thanks!

PR #202, opened 2019-03-17
Avoid deadlock in config_save()

lhca:
Adresses GitHub issue #201.


lfos:
Merged into pu, thanks!

Issue #207, opened 2019-04-16
calcurse-caldav treat reccuring events of the kind "first tuesday of the month" as "each xth day of the month" where x if the day where the event has been created


Psycojoker:
Hello,

Step to reproduce:

- on your caldav server that is not calcurse, create an event like "every first tuesday of the month"
- sync using caldav-calcurse
- open calcurse, inside the first even is good, but the next ones aren't on tuesday but on the xth day of the month that was the day of the created even (like they are always on the 7th instead of tuesday)

I've tried looking a bit at the code calcurse-caldav.py but the bug seems located inside calcurse in "recur.c" and it's been like 10 years since I've done C :/

Kind regards,

lhca:
A recurring event of the type you describe cannot be expressed in calcurse at present.

Calcurse is limited to recurring events which can be expressed as "every _n_-th _period_ starting from _day_", where _n_ is a positive integer, _period_ is either day, week, month or year,  and _day_ is the selected day in the calendar. Examples are "every other day", or "every fourth week", or "every month", all starting from a particular day.

Even that is not without its problems and pitfalls. E.g. what does "every month starting from 31 May" mean? Or "every year starting from 28 February"?

Psycojoker:
Thanks for taking the time to write this detailed answer :)

lhca:
Not at all. As an illustration, try creating a recurring appointment like:

`From 22:30 to 22:45 every month starting on 31 May 2019`

In months with 31 days calcurse will show the appointment correctly. But the recurrence rule also generates an appointment for 31 June which by calcurse is interpreted as

`1 July 2019, 22:30 -> 22:45`

which can be recognized as the day after 30 June. But the iCalendar specification (RFC 5545) requires such recurrence instances must be ignored. Calcurse also displays the following instances which are harder to defend:

`1 March 2020, 22:30 -> ..:..`
`2 March 2020, 22:30 -> 22:45`

lfos:
I agree that this is a bug that should be fixed.

If we want to implement the original request, we should first come up with a way to extend the data file format in a clean and backwards compatible way.

lhca:
The instance
 
`2 March 2020, 22:30 -> 22:45`

is calcurse's interpretation of 31 February 2020, i.e. two days after 29 February 2020 (a leap year). But how is

`1 March 2020, 22:30 -> ..:..`

[...]

PR #208, opened 2019-04-28
Ignore impossible dates

lhca:
Addresses #207.

### DIFF
--- a/src/recur.c
+++ b/src/recur.c
@@ -815,6 +815,18 @@ recur_item_find_occurrence(time_t item_start, long item_dur,
 	lt_item_day.tm_isdst = lt_day.tm_isdst;
 	t = mktime(&lt_item_day);
 
+	/*
+	 * Impossible dates must be ignored (according to RFC 5545). Changing
+	 * only the year or the month may lead to dates like 29 February in
+	 * non-leap years or 31 November.
+	 */
+	if (rpt_type == RECUR_MONTHLY || rpt_type == RECUR_YEARLY) {
+		localtime_r(&t, &lt_item_day);
+		if (lt_item_day.tm_mday != lt_item.tm_mday)
+			return 0;
+	}
+
+	/* Exception day? */
 	if (LLIST_FIND_FIRST(item_exc, &t, exc_inday))
 		return 0;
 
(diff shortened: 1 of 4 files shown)


src/recur.c
```diff
@@ -811,6 +811,17 @@ recur_item_find_occurrence(time_t item_start, long item_dur,
...
+	if (!ISLEAP(1900 + lt_item_day.tm_year) &&
+	    lt_item_day.tm_mon == 1 && lt_item_day.tm_mday == 29)
+		return 0;
+	if (lt_item_day.tm_mon == 1 &&
+	    (lt_item_day.tm_mday == 30 || lt_item_day.tm_mday == 31))
+		return 0;
+	if ((lt_item_day.tm_mon == 3 || lt_item_day.tm_mon == 5 ||
+	     lt_item_day.tm_mon == 8 || lt_item_day.tm_mon == 10) &&
+	    lt_item_day.tm_mday == 31)
+		return 0;
```
lfos: Any reason you are not using check_date() here?

lfos: I guess something like returning 0 if `(rpt_type == RECUR_MONTHLY || rpt_type == RECUR_YEARLY) && !check_date())` (untested) should work.

lfos: Or maybe not, given the way we manipulate the months above. But that makes we wonder whether your check actually works? What happens when using a very high frequency? Say a monthly item starting on February 29th on a leap year and having a frequency of 24 months? Would your check still work?

lhca: I had quite forgotten `check_date()` to be honest. But neither my check nor `check_date()` can work without some tweaking of `lt_item_day.tm_year` and `lt_item_day.tm_mon`. The latter may be negative as you indicated.

Any check should also be limited to the yearly and monthly cases (as weekly and daily lead to no ambiguous dates).

Another thing: I added a test case to the existing `apts-recur` data file. Would you prefer a separate data file?

lfos: What should work is converting the computed `time_t` back to a `struct tm` and verify that the `tm_mday` matches the original `tm_mday`.

Reusing the existing data files in the tests is fine if it does not break any other test cases.

lhca: Indeed! A direct verification is much better.

I have added a test case for yearly and uploaded the changes.

lfos:
Looks great! Merged into pu!

Issue #213, opened 2019-06-01
Inconsistent handling of empty prompts


lfos:
When creating a new appointment and entering an empty end time, the end time is ommited. When specifying the end date of a recurring event, you have to enter `0` to omit the date. Here, the empty string will cancel. I think we should allow to use the empty string to omit the end date (and keep `0`, just for backwards comnpatibility).

Generally, I wonder whether we should cancel processes if a user enters an empty string. It is an invalid input for many prompts and should display a warning, then repeat the prompt.

lhca:
I tried to recreate the scenario you describe, but had quite a different and inexplicable experience.

Looking at the code (which is my own), the reason was obvious: an uninitialized variable. The fix is easy and the result is the desired consistency. There is no question of backwards compatibility; the bug was introduced in 4.4.0.

I have uploaded a fix (PR #214).

lfos:
I tested your patch but the behavior with the patch applied is still exactly the same...

lhca:
I didn't read carefully enough what you wrote and was misled. But I also think that you are mixing things up.

> When creating a new appointment and entering an empty end time, the end time is ommited. When specifying the end date of a recurring event, you have to enter `0` to omit the date. Here, the empty string will cancel. I think we should allow to use the empty string to omit the end date (and keep `0`, just for backwards comnpatibility).

The two cases are quite different. The end _time_ of an appointment cannot be compared with the end _date_ of a recurring event. The latter is better called the until date. It is treated just like the until date of a recurring appointment, which has nothing to do with the end time.

Omitting the end _time_ of an appointment gives you an instant appointment. Omitting the end _date_ gives you an endless recurring appointment (or event), almost the opposite. There is no inconsistency.

The wording is unfortunate and confusing: "end time" and "end date" are too close for comfort. 
They designate two different things which you would never guess from the terms. The solution is to improve the prompt text and use "until" and not "end date" or "duration".

lfos:
If you do not agree it is inconsistent, let's say it is unintuitive instead.

If you want to create an appointment without an end time, you can just enter an empty duration. If you want to create an endless recurring event, you have to enter `0`. If anything, switching these two scenarios would make much more sense: a duration of `0` describes an item without an end time quite well, whereas interpreting the "until date" `0` as infinity seems quite arbitrary. I would just allow the empty string in both cases.

lhca:
Agreed.

lhca:
[...]

PR #214, opened 2019-06-01
Fix recurrent appointment end time when duration is zero

lhca:
Signed-off-by: Lars Henriksen <LarsHenriksen@get2net.dk>

### DIFF
--- a/src/ui-day.c
+++ b/src/ui-day.c
@@ -209,15 +209,10 @@ static void update_duration(time_t *start, long *dur)
 	unsigned newdur;
 	char *timestr, *outstr;
 
-	if (*dur > 0) {
-		end = *start + *dur;
-		asprintf(&outstr, "%s %s", DATEFMT(conf.input_datefmt), "%H:%M");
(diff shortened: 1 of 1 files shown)


lfos:
Since this fixes a regression, I think we should mention the commit that introduced the bug in the commit message (for the record). Is it 223810c (Major overhaul of appointment/event input routines., 2017-11-06)?

lhca:
Yes, in combination with 16d3032 (Refactoring update_duration/day_edit_duration, 2017/10/18).

lfos:
As I mentioned in the bug reports, this does not fix #213. Is this PR just an unrelated fix and #213 still needs work?

lhca:
Clearly there is a bug that is fixed. Whether it is unrelated to #213 I don't know, see my comment there.

lfos:
I am quite confused. Your comment in #213 claims that this PR yields the "desired consistency". Does this not refer to the inconsistency mentioned in the original PR?

lfos:
Merged into pu, thanks!

Issue #220, opened 2019-06-10
Update of recurrent item parameters


lhca:
The recurrent item parameters are type, frequency, until date and (list of) exception dates.

In the master branch version (5283689) the list of exception dates can be edited (like the first three). It is treated separately, which may not be immediately apparent in the UI, but becomes clear when you cancel an edit session (of a recurrent item with exceptions) by pressing ESC while editing type, frequency or until date. The edit session then continues with the exception dates.

This is particularly unfortunate if you cancel while editing the exception dates, because it may lead you to believe that you also cancelled the first part of the edit session.

Instead, the recurrent item parameters should be treated as a unit, and a cancelled edit session should leave all parameters unchanged.

I have a PR coming up.

lfos:
I agree with your analysis and your proposed solution. Looking forward to the PR!

lhca:
> Looking forward to the PR!

The first shot was a bit off the mark. I have uploaded a second attempt.

PR #221, opened 2019-06-10
Treat recurrence item parameters as a unit

lhca:
Adresses GitHub issue #220.

### DIFF
--- a/src/ui-day.c
+++ b/src/ui-day.c
@@ -272,7 +272,29 @@ static void update_desc(char **desc)
 	updatestring(win[STA].p, desc, 0, 1);
 }
 
-static void update_rept(struct rpt **rpt, const long start)
+/* Edit the list of exception days for a recurrent item. */
+static int update_exc(llist_t *exc)
+{
+	int updated = 0;
+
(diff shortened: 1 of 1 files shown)


lfos:
Merged, thanks!

lhca:
Note. Has been updated.

lfos:
I don't see any updates in the branch. What was the problem with the previous attempt?

lhca:
I force-pushed an update (and it is there), but for whatever reason GitHub doesn't say so.

The problem was that the result of recur_update_exc() was not returned and checked. But I have just realized that the also the second attempt is incomplete. Sigh.

lhca:
Third attempt uploaded (force-pushed) as a replacement (not a separate patch).

The problem with the first attempt (still on pu) is that not only the result of `update_exc()`, but also that of  `recur_update_exc()` must be handled. The second attempt did this incorrectly.

lhca:
I have just discovered yet another bug in this error-ridden PR.

If you edit the repetition part of a recurrent item, changes have no effect -- unless the item has an exception date.

A fix is upcoming.

Issue #200, opened 2019-03-14
403 Forbidden when submitting authcode for google calendar


ssbothwell:
I was able to get the auth code successfully, however i have not been able to use it to authorize syncronization:

```
~ λ calcurse-caldav --init keep-remote --authcode '<auth_code_here>'
/usr/bin/calcurse-caldav:563: DeprecationWarning: This method will be removed in future versions.  Use 'parser.read_file()' instead.
  config.readfp(open(configfn))
Connecting to apidata.googleusercontent.com...
Removing all local calcurse objects...
error: The server at apidata.googleusercontent.com replied with HTTP status code
error: 403 (Forbidden) while trying to access https://apidata.googleusercontent.
error: com/caldav/v2/xxx@gmail.com/events/.
```

I'm using `calcurse 4.4.0` on archlinux.

shibacomputer:
I have the same issue on macOS Mojave and am not quite sure how to diagnose it.

lfos:
Did you try to use debug mode?

lfos:
Also, is this possibilty a duplicate of #147?

shibacomputer:
@lfos this is not a duplicate of #147 as this happens at `--init` and not during a sync.

Here's what happens after running calcurse-caldav with a debug flag:

```
$ calcurse-caldav --debug --init keep-remote
/usr/local/bin/calcurse-caldav:563: DeprecationWarning: This method will be removed in future versions.  Use 'parser.read_file()' instead.
  config.readfp(open(configfn))
warning: Dry run; nothing is imported/exported. Add "DryRun = No" to the
warning: [General] section in the configuration file to enable synchronization.
Running command: ['calcurse', '--version']
> REPORT https://apidata.googleusercontent.com/caldav/v2/xxxxxx@xxxxxx.com/events/
> Headers: {'Content-Type': 'application/xml; charset=utf-8', 'Depth': '1'}
> <?xml version="1.0" encoding="utf-8" ?><C:calendar-query xmlns:D="DAV:"                   xmlns:C="urn:ietf:params:xml:ns:caldav"><D:prop><D:getetag /></D:prop><C:filter><C:comp-filter name="VCALENDAR" /></C:filter></C:calendar-query>

< Status: 403 (Forbidden)
[...]

PR #222, opened 2019-06-12
Add Troubleshooting section to calcurse-caldav README.md

watersalesman:
As mentioned in #200. We can collect common issues going forward under this section


lfos:
Awesome, merged into pu. Thanks!

PR #225, opened 2019-06-26
Recurrence rule correction

lhca:
I came across the following while trying to extend support for RFC 5545 (iCalendar) recurrence rules.

Try the following illustrative experiment in the multiple days interface:

- create an appointment on 31 December 2019, 12:00 that ends on 1 January 2020, 11:00
- turn it into a repeating appointment of type yearly, frequency 2, endless (no until date), and watch closely when you type the last `RETURN`
- step one year forward one or several times and watch out

The PR includes a patch and test cases.


lfos:
Good catch. Merged into pu, thanks!

lhca:
On the pu branch the new test case `recur-007.sh` fails.

The reason is that by mistake the commit https://github.com/lfos/calcurse/pull/225/commits/d29c1c660c42b38b435b5e19d91f7f7e41283f9d contained a wrong version. That can happpen, but why was it not picked up by Travis?

I have uploaded an update (https://github.com/lhca/calcurse/commit/f7c79e0cbe822901d3821189d12bb9e64c531be9) to the original pull-request branch.

PR #226, opened 2019-07-09
Fix edit of empty exception day list

lhca:
Correction to #221.

If the exception day list is left untouched, it has been correctly "updated". This is also true of the empty list.


lfos:
Merged into `pu`, thanks!

Issue #228, opened 2019-07-18
[Feature Request] Create Appointment from TODO


igordsm:
Creating an appointment from a TODO item would be very convenient to me, as I like to track time spent on each TODO. It would also help me to organize my schedule by assigning a specific part of day to do each task. 

Would this be a useful addition to calcurse? For me this would be useful even if appointments and todos are not linked together (i.e. just creating a separate and completely independent appointment with the same text as the source TODO item).

Best regards

PR #229, opened 2019-07-20
Feature Request #228: create appointment from task

igordsm:
Implementation of Issue #228, which requests the addition of an option to create Appointments from the TODO panel with the text of the current task as the appointments' description.


igordsm:
Hey @lfos, could you review this when you have some time? I don't want to put pressure on you or anything, but I've been using this PR for a week and it would be awesome if it was included in the official releases!

lfos:
Of course, thanks for the PR. The patch looks good. The only thing I don't like is using one of the few keys that we have left. If we continue doing this, we will run out of keys at some point and there will be no way back. I think we should switch to vim-like key prefixes for some new (less common commands). For example, we could use `za` to turn a todo into an appointment.

igordsm:
@lfos  thank you! I added a new commit to the PR removing the Z shortcut and adding a commando to key_generic_cmd. If things look good I can squash it to a single commit to make life easier for you. 

Best regards.

lfos:
I was not talking about turning it into a command. Why I suggested is being able to press `z` followed by `a` in normal mode to generate the new item. This allows us to add further key bindings prefixed by `z` later. For example, `zj` could jump to the next appointment in the appointments panel.

igordsm:
Oh, I guess I was confused about the correct naming of thing in calcurse. Sorry about that. You mean like the "C" key for Config ?

lfos:
> Oh, I guess I was confused about the correct naming of thing in calcurse. Sorry about that. You mean like the "C" key for Config ?

Exactly!

kevinhwang91:
@igordsm Adding todo to appt, the description works fine but not the notes.

igordsm:
@kevinhwang91 I have a prototype that also added the note to the appointment.  Unfortunately I have not found a way to link the appointment's note to the original Task's. I can copy it, but changing one does not automatically change the other. 

I've been a bit busy this past week, but I plan to come back to calcurse by the end of this week.

lfos:
Yes, there currently isn't any code that allows to attach the same note to multiple items (and keep the content synced).

A similar idea that has been discuessed multiple times in the past is named notes. Think of it as a symlink-like indirection: an appointment points to a note name and the note name points to the actual note. If you edit the note, it is still available under the original note name and all references to that named note will automatically point to the new note content.

lfos:
What's the status of this? To avoid more bikeshedding, maybe we should simply create a copy of the note for now?

lfos:
Closing for now. Please reopen if you find time to complete this patch set!

Issue #169, opened 2019-01-16
[Feature Request] adding todo-s in batch


pearupung:
When doing the planning phase for the next day, it is quite tiresome to press Ctrl+T every time you think of a new todo. It would be nice to press some key combination (Ctrl+Shift+T for example) to enter "todo adding mode" which would just loop the two steps of creating a todo and exit the mode if 'q' is pressed.

lfos:
I like this idea!

igordsm:
How about an "Add Batch" function that would call "Add Item" until an addition is cancelled? This could work with Appointments and Events as well.

PR #230, opened 2019-07-21
Adds batch add functionality requested in #169

igordsm:
This PR adds an option "Add Batch" (shortcut "b B") that keeps calling "key_add_item" until the addition is cancelled by pressing "ESC". It works for both tasks and appointments/events. Requested in issue #169


igordsm:
@lfos You mentioned in #229 that we should avoid adding new keys. Do you think we should apply it to this command as well?

lfos:
Are there other operations where a batch mode might make sense? Batch updates, batch removals? If so, we might want to make `b` a prefix key that can be used with other operations.

lhca:
The global flag `modified` in `io.c` mirrors the state of online (in-memory) data as opposed to offline (on-disk) data: 

- when it is unset, in-memory and on-disk data are in agreement.
- when it is set, in-memory data disagree with on-disk data

The flag is unset at start of calcurse and after each successful save operation.

Your patch breaks this mechanism in a threaded application like calcurse which has a periodic save thread.

igordsm:
@lhca I did not realize this! I will update the patch later this week to fix this.

igordsm:
@lhca I refactored the code so that `key_add_item` (and the functions it calls) returns 1 if the element really was added. This avoids using the `modified` flag and makes the rest of the patch trivial.

lhca:
Yes, giving `key_add_item()` a return value seems the obvious and simple answer.

igordsm:
@lhca I squashed both commits. 

@lfos Thinking again, I don't think the suggestion applies here. 

Any other suggestions for improvements?

lfos:
Any updates on this. Did you think about making the batch mode key a prefix key as discussed previously?

igordsm:
Hey @lfos , sorry for the wait.  I initially implemented it using the Add Item command, but it does not allow for different types of adding in the same batch. Do you think adding a batch mode key prefix would be useful? The "next level" of menu would contain the type of the item to add, in this case. 

I am not sure I would use, but am willing to implement it if others think it would be useful.

igordsm:
Hey @lfos , sorry for the wait.  I initially implemented it using the Add Item command, but it does not allow for different types of adding in the same batch. Do you think adding a batch mode key prefix would be useful? The "next level" of menu would contain the type of the item to add, right?

I am not sure I would use, but am willing to implement it if others think it would be useful.

lfos:
Sorry again for the super late reply. We should work on getting this done. I don't think the second level should only allow you to select an item type. We can simply allow for choosing some of the already existing key bindings. As a first step, I'd be fine with only supporting `a` (add item), `^A` (add appointment) and `^T` (add todo item).

lfos:
Closing for now. Please reopen if you find time to complete this patch set!

Issue #18, opened 2016-12-06
CalCurse-CalDav: Cannot handle HTTP connections


bbenne10:
At work, I am currently tied to an Exchange 2010 environment and thus am using DavMail to patch over the silliness. This exposes a CalDav server on port 1080. On this port, DavMail translates traffic to EWS and all is well.

However, calcurse-caldav cannot cope with this due to the assumption that all CalDav traffic is https. I have patched over this locally by changing line 508 to `conn = http.client.HTTPConnection(hostname)`, but it'd be nice to have some built-in support for this. A configuration variable might be a good idea, as we already have one for `InsecureSSL`. Thoughts?

lfos:
Yes, a configuration value sounds good to me but it should default to
HTTPs. Maybe `UseSSL`? Or just `SSL`?

bbenne10:
Either works for me. SSL feels more consistent internally. 

On a second thought, maybe something like below would work? This would replace ~481 - 496. The inclusion of a scheme feels much more natural to me (though admittedly the scheme parsing is pretty weak when compared to `urllib.urlparse` - which feels like the right way to get all this information rather than asking for `HostName`, `Path` and `SSL` - but I didn't want to deviate TOO far for this one issue. I can open another shortly if we want to talk about that too)

```python
context = None # noop, as this is default value for http.client.HTTPSClient
if insecure_ssl:
    try:
        context = context = ssl._create_unverified_context()
    except AttributeError:
        pass

# we might want "hostname" to have a different name, as it's now a server uri
hostname = hostname.lower()
ssl = True
if hostname[0:6] == 'http://':
    hostname = hostname[0:6]
    ssl = False
elif hostname[0:7] == 'https://':
    hostname = hostname[0:7]

if ssl:
    conn = http.client.HTTPSConnection(hostname, context=context)
else:
    conn = http.client.HTTPConnection(hostname)
```

lfos:
You can now turn off HTTPs using the `HTTPS` option in the configuration file.

PR #234, opened 2019-08-18
Document the HTTPS option in the sample configuration file

k0ral:
#18 introduced the `HTTPS` option but it is missing from the sample configuration file, therefore it is not easily discoverable by new users (like me.)


lfos:
Thanks! Maybe we should add a note to the comment, mentioning that using plan HTTP is highly discouraged?

k0ral:
Comment updated.

lfos:
Merged into pu, thanks!

PR #235, opened 2019-08-26
Until day inconsistencies

lhca:
There are some inconsistencies in the way the end day (until day) of a recurrent item is handled. The problems are minor and only visible in interactive use of calcurse as in the following scenario:

1) Create an appointment with start time 12:00, end time 11:59 (2-day appointment)

2) Turn it into a recurrent appointment of type daily, frequency 3, until day +3

The appointment is correctly displayed with two 2-day occurrences, three days
apart.

3) Edit the appointment and select Repetition. Accept existing type, frequency and end day (now as a date)

The second day of the second occurrence disappears.

4) Repeat 3), but set the end day as an offset (+3)

The second day of the second occurrence reappears. It will also be there if calcurse is restarted (data are loaded), but you can make it disappear by repeating 3).

For an explanation, see the commit message where the scenario is also described.

### DIFF
--- a/src/recur.c
+++ b/src/recur.c
@@ -433,8 +433,8 @@ struct recur_apoint *recur_apoint_scan(FILE * f, struct tm start,
 	tend = mktime(&end);
 
 	if (until.tm_year != 0) {
-		until.tm_hour = 23;
-		until.tm_min = 59;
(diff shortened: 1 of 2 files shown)


lfos:
Merged into pu, thanks!

PR #237, opened 2019-09-09
Fix German translation for deleting one repeating item

NWuensche:
As the hotkey for deleting only one occurrence in german is not 'o', but
'e', the tip is misleading. Fix it by highlighting 'e' now.


NWuensche:
I tried writing to "bugs@calcurse.org" which I found in `de.po` but I don't have permissions to write there.  Also, the `Language-Team` URL "http://www.transifex.com/lfleischer/calcurse/language/de" does not exist. What should I do now?

Issue #137, opened 2018-08-08
calcurse-caldav does not create sync.db


marskar:
First of all, thank you for calcurse, I really love it, though I do have a small problem with the `calcurse-caldav` script. 

The script does not create `sync.db`, so I have to run `calcurse-caldav --init keep-remote` instead of just `calcurse-caldav` each time.
To be clear, `~/.calcurse/caldav/` only contains `config` and `oauth2_cred`, not `sync.db`.

I am using MacOS with python 3.6.5.
The only error I get is after importing new objects: `error: Missing ETag`.
Happy to share more info, but I don't think the issue is with my `config` file.

I did change the shebang in the script to use my anaconda python interpreter:
```python
#!/Users/skarzynskimw/anaconda3/bin/python

import argparse
import base64
import configparser
import httplib2
import os
import re
import subprocess
import sys
import textwrap
import urllib.parse
import xml.etree.ElementTree as etree

# Optional libraries for OAuth2 authentication
try:
    from oauth2client.client import OAuth2WebServerFlow, HttpAccessTokenRefreshError
    from oauth2client.file import Storage
    import webbrowser
except ModuleNotFoundError:
    pass


def msgfmt(msg, prefix=''):
    lines = []
    for line in msg.splitlines():
        lines += textwrap.wrap(line, 80 - len(prefix))
    return '\n'.join([prefix + line for line in lines])


def warn(msg):
    print(msgfmt(msg, "warning: "))


def die(msg):
    sys.exit(msgfmt(msg, "error: "))


def die_atnode(msg, node):
    if debug:
        msg += '\n\n'
        msg += 'The error occurred while processing the following XML node:\n'
        msg += etree.tostring(node).decode('utf-8')
    die(msg)


def calcurse_wipe():
    if verbose:
        print('Removing all local calcurse objects...')
    if dry_run:
        return
    subprocess.call([calcurse, '-F', '--filter-hash=XXX'])


def calcurse_import(icaldata):
    p = subprocess.Popen([calcurse, '-i', '-', '--dump-imported', '-q',
                          '--format-apt=%(hash)\\n',
                          '--format-recur-apt=%(hash)\\n',
                          '--format-event=%(hash)\\n',
                          '--format-recur-event=%(hash)\\n',
                          '--format-todo=%(hash)\\n'],
                         stdin=subprocess.PIPE, stdout=subprocess.PIPE)
    return p.communicate(icaldata.encode('utf-8'))[0].decode('utf-8').rstrip()


[...]

PR #243, opened 2019-09-24
<#137> Handle objects with status code 404 aborting db syncing

zzandland:
I was having the same issue as everyone else in https://github.com/lfos/calcurse/issues/137. This PR is exactly the same code as in https://github.com/lfos/calcurse/issues/137#issuecomment-490541349. His solution works perfectly fine but I don't know why he decided not to make any PR for the solution. So I am making it instead.

### DIFF
--- a/contrib/caldav/calcurse-caldav.py
+++ b/contrib/caldav/calcurse-caldav.py
@@ -3,7 +3,6 @@
 import argparse
 import base64
 import configparser
-import httplib2
 import os
 import re
 import subprocess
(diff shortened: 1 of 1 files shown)


lfos:
Thanks for submitting a PR! Could you please squash the second patch into the first one to make it easier to review your proposed changes?

lfos:
I squashed the patches, added a more detailed commit message and pushed the patch to `pu`. Thanks again for the PR!

PR #254, opened 2019-12-17
Add XDG Base Directory support

### LINKED ISSUE #252: [Feature Request] XDG Base Directory Support
calcurse stores its files by default in `$HOME/.calcurse`.

It should store config files in `$XDG_CONFIG_HOME/calcurse` and its data files in `$XDG_DATA_HOME/calcurse` instead.

Defaults if not set:
- `XDG_CONFIG_HOME`: `"$HOME/.config"`
- `XDG_DATA_HOME`: `"$HOME/.local/share"`

Current workaround:
```shell
$ calcurse -C "$XDG_CONFIG_HOME/calcurse" -D "$XDG_DATA_HOME/calcurse"
```

Alternatively, keep `$HOME/.calcurse` as default but if `$HOME/.config/calcurse` or `$HOME/.local/share/calcurse` exist then use those instead.

Nitroretro:
This PR changes calcurse's default file structure from this:

```
$HOME/.calcurse/
          |___apts
          |___conf
          |___hooks/
          |___keys
          |___notes/
          |___todo
```

To this:
```
$XDG_DATA_HOME/calcurse/        $XDG_CONFIG_HOME/calcurse/
                  |___apts                          |___conf
                  |___notes/                        |___hooks/
                  |___todo                          |___keys
```

**Defaults:**
* `$XDG_DATA_HOME`: `$HOME/.local/share`
* `$XDG_CONFIG_HOME`: `$HOME/.config`

In addition, I have updated all documentation to reflect this change and have made `io_check_dir` create parent directories as well, like `mkdir -p`. This is because if `$HOME/.local/share` doesn't exist we have to make sure `$HOME/.local` is created as well.

If `$HOME/.calcurse` exists, then it will use the old file structure instead for backward compatibility.

I have also edited the `calcurse-caldav.py`, `calcurse-vdirsyncer` scripts and the `pre-load` hook to use the XDG directories, but **I haven't edited** the `post-save` and `post-sync` hooks since they require `apts`, `conf`, `keys` and `todo` to all be in the same directory. How should that be handled?

Fixes #252.

P.S: Thanks for calcurse, it's great :)

### DIFF
--- a/doc/calcurse.1.txt
+++ b/doc/calcurse.1.txt
@@ -95,7 +95,7 @@ subsections contain some general desriptions of command line options and usage.
 
 Input and Output Date Format
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-Many options require a 'date' argument, and query results per day are set apart 
+Many options require a 'date' argument, and query results per day are set apart
 by a leading 'date line'.
 
 The input format of 'date' options and the output format of 'date lines' are
@@ -143,19 +143,22 @@ are marked "('also interactively')".
 
 *-c* 'file', *--calendar* 'file'::
   ('also interactively') Specify the calendar file to use. The default
-  calendar is *~/.calcurse/apts* (see <<_files,FILES>>). If 'file' is not an
-  absolute path name, it is interpreted relative to the current working
-  directory. The option has precedence over *-D*.
+  calendar is *$XDG_DATA_HOME/apts* (*\~/.local/share/calcurse/apts*) or
+  *\~/.calcurse/apts* if *~/.calcurse* exists.  (see <<_files,FILES>>). If
+  'file' is not an absolute path name, it is interpreted relative to the current
+  working directory. The option has precedence over *-D*.
 
 *-C* 'dir', *--confdir* 'dir'::
   ('also interactively') Specify the configuration directory to use. If not
-  specified, the default directory is *~/.calcurse/*. See <<_files,FILES>> for
-  the interaction with *-D*.
+  specified, the default directory is *$XDG_CONFIG_HOME/calcurse/*
+  (*\~/.config/calcurse/*) or *~/.calcurse* if it exists.  See <<_files,FILES>>
+  for the interaction with *-D*.
 
 *-D* 'dir', *--datadir* 'dir'::
   ('also interactively') Specify the (data) directory to use. If not specified,
-  the default directory is *~/.calcurse/*.  See section <<_files,FILES>> for
-  the interaction with *-C*.
+  the default directory is *$XDG_DATA_HOME/calcurse*
+  (*\~/.local/share/calcurse/*) or *~/.calcurse* if it exists.  See section
+  <<_files,FILES>> for the interaction with *-C*.
 
 *-d* 'date|num', *--day* 'date|num'::
   Print appointments and events for the given date or given range of days,
@@ -622,14 +625,14 @@ The following structure is created by default in your home directory the
 first time calcurse is run without any options:
 
 ----
-$HOME/.calcurse/
-          |___apts
-          |___conf
-          |___hooks/
-          |___keys
-          |___notes/
-          |___todo
+$XDG_DATA_HOME/calcurse/        $XDG_CONFIG_HOME/calcurse/
+                  |___apts                          |___conf
+                  |___notes/                        |___hooks/
+                  |___todo                          |___keys
 ----
+
++$XDG_DATA_HOME+ defaults to +\~/.local/share+ and +$XDG_CONFIG_HOME+ defaults to
++~/.config+.
 
 The files are of two different kinds: data and configuration. The data files
 constitute the calcurse database and are independent of the calcurse release
@@ -661,14 +664,18 @@ scripts, see <<_hooks,Hooks>>.
 Directory configuration
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-An alternative directory to the default +$HOME/.calcurse+ may be specified
-with the *-D* option.
+An alternative directory to the defaults +$XDG_DATA_HOME/calcurse+
+(+$HOME/.local/share/calcurse+) and +$XDG_CONFIG_HOME/calcurse+
+(+$HOME/.config/calcurse+) may be specified with the *-D* option.
 
 An alternative directory for the configuration files 'only' may be specified
 with the *-C* option; in that case data files are either in the default
 directory or in the directory specified by *-D*. If both *-D* and
 *-C* are present, configuration files in the data directory, if any, are
 ignored.
+
+If +$HOME/.calcurse+ exists, then it will be used as the default for both the
+data directory and the configuration directory.
 
 ----
 <datadir>      <confdir>
@@ -677,7 +684,10 @@ ignored.
      |__ todo      |___ keys
      |__ notes/    |___ hooks/
 
-default for both: $HOME/.calcurse/
+defaults:
+     <datadir>: $XDG_DATA_HOME/calcurse ($HOME/.local/share/calcurse)
+     <confdir>: $XDG_CONFIG_HOME/calcurse ($HOME/.config/calcurse)
+     both: $HOME/.calcurse (only if it exists)
 ----
 
 calcurse may switch between two configuration setups, but still access
@@ -686,15 +696,16 @@ the same data files e.g. with:
 ----
 $ calcurse
 
-$ calcurse -C "$HOME/.calcurse/config"
+$ calcurse -C "$HOME/.config/calcurse/config"
 ----
 
 Hooks
 ~~~~~
 
-Scripts placed in +$HOME/.calcurse/hooks/+ trigger actions at certain
-events. To enable a hook, add a script with one of the following names to this
-directory. Also make sure the script is executable.
+Scripts placed in +$XDG_CONFIG_HOME/calcurse/hooks/+
+(+$HOME/.config/calcurse/hooks+) trigger actions at certain events. To enable a
+hook, add a script with one of the following names to this directory. Also make
+sure the script is executable.
 
 *pre-load*::
   Executed before the data files are loaded.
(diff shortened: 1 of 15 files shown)


src/io.c
```diff
@@ -143,15 +143,35 @@ unsigned io_fprintln(const char *fname, const char *fmt, ...)
...
-		asprintf(&path_ddir, "%s%s", datadir, "/" DIR_NAME);
-	} else
+	char* home_dir = getenv("HOME");
+
+	if (!datadir)
+		if (home_dir) {
+			asprintf(&path_ddir, "%s%s", home_dir, "/" DIR_NAME_LEGACY);
+			if (!io_dir_exists(path_ddir)) {
+				if ((path_ddir = getenv("XDG_DATA_HOME")))
+					asprintf(&path_ddir, "%s%s", path_ddir, "/" DIR_NAME);
```
lfos: I think this creates a memory since you do not free the memory you allocated with the previous asprintf() call.

src/io.c
```diff
@@ -143,15 +143,35 @@ unsigned io_fprintln(const char *fname, const char *fmt, ...)
...
+				else
+					asprintf(&path_ddir, "%s%s", home_dir, "/.local/share/" DIR_NAME);
+			}
+		} else {
+			path_ddir = "./." DIR_NAME;
```
lfos: Is this safe? We assign allocated memory to `path_ddir` everywhere else.

src/io.c
```diff
@@ -143,15 +143,35 @@ unsigned io_fprintln(const char *fname, const char *fmt, ...)
...
+	else
 		asprintf(&path_ddir, "%s%s", datadir, "/");
 
 	if (!confdir)
-		asprintf(&path_cdir, "%s%s", path_ddir, "/");
+		if (home_dir) {
+			asprintf(&path_cdir, "%s%s", home_dir, "/" DIR_NAME_LEGACY);
+			if (!io_dir_exists(path_cdir)) {
+				if ((path_cdir = getenv("XDG_CONFIG_HOME")))
+					asprintf(&path_cdir, "%s%s", path_cdir, "/" DIR_NAME);
```
lfos: Potential memory leak. See above.

doc/calcurse.1.txt
```diff
@@ -143,19 +143,21 @@ are marked "('also interactively')".
...
+  (see <<_files,FILES>>).  If 'file' is not an absolute path name, it is
+  interpreted relative to the current working directory. The option has
+  precedence over *-D*.
 
 *-C* 'dir', *--confdir* 'dir'::
   ('also interactively') Specify the configuration directory to use. If not
-  specified, the default directory is *~/.calcurse/*. See <<_files,FILES>> for
-  the interaction with *-D*.
+  specified, the default directory is *$XDG_CONFIG_HOME/calcurse/*
+  (*~/.config/calcurse/*). See <<_files,FILES>> for the interaction with *-D*.
```
lfos: We should definitely mention that if `~/.calcurse/` exists, then it takes precedence over `~/.config/calcurse/`. Same goes for the other options. Also, while rebasing, it would be great if you could change the commit subjects to use imperative mood ("Change documentation to [...]").

src/io.c
```diff
@@ -162,16 +162,20 @@ void io_init(const char *cfile, const char *datadir, const char *confdir)
...
+			if (home_dir) {
+				asprintf(&path_cdir, "%s%s", home_dir, "/" DIR_NAME_LEGACY);
+				if (!io_dir_exists(path_cdir)) {
+					if ((path_cdir = getenv("XDG_CONFIG_HOME")))
+						asprintf(&path_cdir, "%s%s", path_cdir, "/" DIR_NAME);
+					else
+						asprintf(&path_cdir, "%s%s", home_dir, "/.config/" DIR_NAME);
+				}
+			} else {
+				path_cdir = "./." DIR_NAME;
```
lfos: Shouldn't this whole change be squashed into the first patch that touches io_init()? Also need to address memory leaks, as pointed out in my review of the other patch.

doc/calcurse.1.txt
```diff
@@ -624,10 +624,10 @@ The following structure is created by default in your home directory the
...
 
 ----
-$XDG_DATA_HOME/calcurse/    $XDG_CONFIG_HOME/calcurse/
-          |___apts                    |___conf
-          |___notes/                  |___hooks/
-          |___todo                    |___keys
+$XDG_DATA_HOME/calcurse/        $XDG_CONFIG_HOME/calcurse/
+                  |___apts                          |___conf
+                  |___notes/                        |___hooks/
+                  |___todo                          |___keys
```
lfos: This looks quite strange, at least in the GitHub source code review. Could you clarify why this is needed/better?

Nitroretro: It looks like the way it should for me in the GitHub source code review. The reason I added that was because I thought the vertical line should start from `calcurse` instead of the parent directory, like it does in the original one.

Nitroretro:
@lfos I believe everything should be good now, this was my first time programming in C and I didn't realize I had to free the memory, my bad.

lfos:
No worries. This is actually pretty good code, considering that it's your first time programming in C!

I made some very small changes (you forgot to remove an asprintf() call and some commit messages were way too long) and pushed the commits to `pu`.

Thanks again. Great job!

Nitroretro:
Oops, you're right. I remember deleting it but somehow I managed to put it back in. It's my first time programming in C, but I've dabbled a bit in C++ and I've been programming for a decent amount of time.

lfos:
It might be worthwhile adding some test cases to verify that the precedence of directories works as intended.

lhca:
I have now had an opportunity to read the updated man page and have some comments.

A major change is that data and configuration files are now _by default_ in two different directories. The previous default was one directory, and this directory could be changed with the "-D" option _alone_. This is no longer so. The "-D" option now changes only the data directory, but this is not reflected properly in the man page. The description of the "-D" option is inaccurate, and the first and second paragraphs of section 8.3, Directory configuration, are erroneous.

As a matter of fact, the changes to "-D" means that patch as a whole is not backwards compatible.

The man page has an `ENVIRONMENT` section. Doesn't `XDG_DATA_HOME` and `XDG_CONFIG_HOME` belong here with an explanation of their meaning, use and default values if not set? Mention of the default values everywhere (in parentheses) makes for heavy reading and is confusing. Refer to `ENVIRONMENT` instead.

The user home directory should be referred to as "$HOME" throughout and not sometimes as "~".

In the explanation of the "-c" option, "The default calendar is $XDG_DATA_HOME/apts ..." should be "The default calendar is $XDG_DATA_HOME/calcurse/apts ...".

In the `FILES` section, the phrase "in your home directory" should be left out.

lfos:
Thanks for the comments, Lars! I am not too convinced we need to mention `XDG_DATA_HOME` or `XDG_CONFIG_HOME` in the `ENVIRONMENT` section -- it's not a calcurse-specific environement variable and we also don't mention `HOME` there.

I agree with all you other comments, though, and the backwards-incompatible change of `-D` is something we should definitely address. I like the "new" behavior better but I think it might be an option that is used quite heavily, so we need to find a good solution that doesn't suddently break a lot of setups.

Nitroretro:
Thanks for the comments!

---

> The "-D" option now changes only the data directory, but this is not reflected properly in the man page.

The "-D" option still changes both the data and the configuration directories:

https://github.com/lfos/calcurse/blob/0754cbc9e03d9fabfaacd4d5e0649b48b77390e3/src/io.c#L170-L174

The `-C` option still takes priority, but if it's not specified then it will check if `-D` was specified and use that instead. I kept this behavior since we are aiming for backward compatibility.

---

> Mention of the default values everywhere (in parentheses) makes for heavy reading and is confusing. Refer to ENVIRONMENT instead.

I suggest we refer to `FILES` everywhere instead of specifying the default value and also mention `FILES` inside the `ENVIRONMENT` section (example: `See also section FILES.`). Referring to `ENVIRONMENT` would mean having to explain the default values twice, once in `FILES` and once in `ENVIRONMENT`, and I think `FILES` fits it better. The only time I've seen `XDG_CONFIG_HOME` or `XDG_DATA_HOME` being mentioned in environment was in `systemd(1)`.

---

> The user home directory should be referred to as "$HOME" throughout and not sometimes as "~".

I agree, this inconsistency existed before as well.

---

> In the explanation of the "-c" option, "The default calendar is $XDG_DATA_HOME/apts ..." should be "The default calendar is $XDG_DATA_HOME/calcurse/apts ...".
> In the FILES section, the phrase "in your home directory" should be left out.

You're right, nice catch!

---

Should I create a new pull request with these changes?

lhca:
I withdraw most of my criticism.

> > The "-D" option now changes only the data directory, but this is not reflected properly in the man page.
> 
> The "-D" option still changes both the data and the configuration directories:
> 
> The `-C` option still takes priority, but if it's not specified then it will check if `-D` was specified and use that instead. I kept this behavior since we are aiming for backward compatibility.

My apologies. I didn't read the code carefully enough. You have indeed managed backward compatibility. Nice work.

So with `-D` alone, data and configuration files are kept in one place as before and in contrast to the default. Admitted, the first paragraph of section 8.3 is accurate as it stands, but I think this point should be stressed explicitly.
 
> > Mention of the default values everywhere (in parentheses) makes for heavy reading and is confusing. Refer to ENVIRONMENT instead.
> 
> I suggest we refer to `FILES` everywhere instead of specifying the default value and also mention `FILES` inside the `ENVIRONMENT` section (example: `See also section FILES.`). Referring to `ENVIRONMENT` would mean having to explain the default values twice, once in `FILES` and once in `ENVIRONMENT`, and I think `FILES` fits it better. The only time I've seen `XDG_CONFIG_HOME` or `XDG_DATA_HOME` being mentioned in environment was in `systemd(1)`.

Yes, `FILES` is better.

lfos:
> Should I create a new pull request with these changes?

The patch series is still in `pu`, so I can squash any fixes into the original patches. Feel free to open a new PR. It would be nice to split fixes that should be added to separate commits.

PR #255, opened 2019-12-21
Test for the directory structure

Nitroretro:
Ensure that calcurse uses the right data and configuration directories depending on the conditions.

**Related:** #254 (issue: #252)

### DIFF
--- a/test/Makefile.am
+++ b/test/Makefile.am
@@ -4,6 +4,12 @@ TESTS = \
 	true-001.sh \
 	run-test-001.sh \
 	run-test-002.sh \
+	io-001.sh \
+	io-002.sh \
+	io-003.sh \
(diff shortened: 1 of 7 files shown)


lfos:
Awesome, thank you! Merged into `pu`.

Nitroretro:
No problem!

lhca:
Older Linux installations (mine for example) do not have `realpath`, which causes `io-005.sh` and `io-006.sh` to fail. But `readlink -f` will do.

lfos:
> Older Linux installations (mine for example) do not have `realpath`, which causes `io-005.sh` and `io-006.sh` to fail. But `readlink -f` will do.

Good catch! I updated the patch on `pu`.

Issue #239, opened 2019-09-11
Recurrence data lost when importing ics data


jeromg:
**Version:** 4.5.0

**Action:** import an ICS file which includes a weekday recurrence  
  
**Actual result:** event with a weekly recurrence on Mondays  
  
**Expected result:** recurrence info is preserved for all 5 weekdays  
  
**Repro**  
1. Import the following ICS event using calcurse -i [filename]  
```
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//acme//NONSGML remmy v0.1
METHOD:PUBLISH

BEGIN:VEVENT
ORGANIZER:MAILTO:me@machine
UID:6D68F3FD-D4D3-11E9-B96B-CAB4B35793DA
LOCATION:see description
CLASS:PUBLIC
RRULE:FREQ=WEEKLY;WKST=SU;BYDAY=MO,TU,WE,TH,FR
DTSTART:20190909T083000
DTEND:20190909T083000
SUMMARY: Check email
DESCRIPTION: Check email
DTSTAMP:20190911T223335
END:VEVENT

END:VCALENDAR
```
  Note: this ICS file is generated from a [remind ](https://dianne.skoll.ca/projects/remind/) file using [remmy.pl](https://github.com/smartmic/remmy)

2. In calcurse I only see a weekly event, every Monday morning at 8:30am, nothing for the other weekdays  

3. If I export the data using: calcurse -xical, I get:  
```
BEGIN:VCALENDAR
PRODID:-//calcurse//NONSGML v4.5.0//EN
VERSION:2.0
BEGIN:VEVENT
DTSTART:20190909T083000
RRULE:FREQ=WEEKLY;INTERVAL=1
SUMMARY:Check email
END:VEVENT
END:VCALENDAR
```

lfos:
`BYDAY` recurrence rules are not yet supported in calcurse, see #207 for details.

jeromg:
Cristal clear, thanks for your prompt answer, I'll keep using calcurse anyway. I was just tinkering for some a new use case.  
  
Cheers  /jerome

Issue #173, opened 2019-01-24
More complicated repeating events


chiraag-nataraj:
I know you can create simple repeating events (events which repeat on a daily, weekly, monthly, yearly, etc) basis, but right now, there doesn't seem to be any easy way to create an event that e.g. occurs every Tuesday and Thursday. Is there something I'm missing, or is this currently not possible?

lfos:
Create the event on the first Tuesday, choose weekly repetition. Copy the event on the first Thursday.

chiraag-nataraj:
That is, in fact, what I ended up doing. But since the events aren't linked, it means that if I need to edit the description at some point, I then have to edit two different events rather than just the one (and having it propagate through all repeats).

chiraag-nataraj:
Is there a possiblity of allowing linking of events? That way, it's a more generic approach which would also allow what I'm describing here without (hopefully) significantly complicating the syntax. Every event could be given an ID and events with the same ID are tied together (so editing will affect all of them). In terms of the syntax, this would just involve adding a field (maybe at the beginning). In terms of upgrades, nothing would be done automatically, but instead would generate IDs the first time you use the link feature (that way, backwards-compatibility is conserved).

hiqua:
I think the PR linked solves this issue, and it's now released since 4.7.0.

PR #258, opened 2020-01-14
Recurrence rule extensions

lhca:
Repeating events beyond the basic ones (daily, weekly, monthly and yearly) have been requested in several issues (#207, #173, #239).

This PR implements parts of the iCalendar (RFC5545) recurrence rules, namely BYDAY, BYMONTHDAY and BYMONTH. The RFC is a somewhat daunting document, but a couple of examples will illustrate what is possible: a weekly event that occurs on Tuesday and Thursday (#173), a weekly event that occurs on working days only (#239), an event that occurs on the first Tuesday of the month (#207).

Test data are in test/data/rfc5545 and include all examples from RFC5545, section 3.8.5.3, which are supported (23), and four others. The event/appointment descriptions contain the text from the RFC describing the repetition pattern as well as the RRULE.

Here is how you interactively set up a repeating event for start of daylight-saving time. The first three steps create a repeating event in the wellknown fashion.

1. Start calcurse and go to the first day on which the event should appear (in the EU: last Sunday in March). Assume March 29, 2020.

1. Create an event "Summer time begins" ('a', no start time).

1. Turn it into a repeating event ('r'): type yearly, frequency 1, duration 0.

1. Edit it ('e') and choose 2 (Repetition). Accept what you just entered for type, frequency and duration and (NEW!):
 - for Weekdays enter '-1Sun' (i.e. the last Sunday)
 - for Months enter '3' (i.e. March)
 - for Monthdays just type return 

Note that instead of 'Sun' you may have to enter the weekday abbreviation of your locale as shown in the calendar panel.

In the apts data file the event will appear as the line:

03/29/2020 [1] {1Y w-7 m3} Summer time begins

For an event that occurs on Tuesday and Thursday, create a repeating weekly event with frequency one that starts on a Tuesday (or Thursday). Then edit it and set Weekdays to 'Tue Thu'. The event type could also have been set to daily with the same result. By studying the examples in the test file mentioned above you can (hopefully) figure out what to input for Weekdays, Months or Monthdays in other cases.

For an explanation of the extended repetition format, see the commit message in "Recurrence rule extensions". 

The PR consists of seven patches (commits). The first four mostly contain necessary refactoring for the recurrence rule extensions in the fifth. Note that, as yet, there is no ICAL import support.


lfos:
First of all, thanks a lot for all the great work! I would really like to see support for extended recurrence rules land in calcurse and this looks really promising! I did not yet have a detailed look at the individual patches but applied the patch series locally and gave it a quick try. Before I start digging deeper, I have some questions and comments on the UI.

Is there any reason extended repetitions are not offered when creating a repetition? Is adding a repetition, then editing it the only way to make use of this feature (without importing events)? On a related note, I found the prompts for editing repetitions highly confusing and wouldn't know how to use them without reading parts of the code or commit messages. I hope we can come up with something that is more intuitive. I'd also be happy if we could keep the process pretty short for "simple" recurrence rules and only show more prompts if the user tells us he wants something more sophisticated. We already ask for the repetition type when adding/editing a repetition. Maybe add one or multiple options for extended repetitions?

lhca:
> Is there any reason extended repetitions are not offered when creating a repetition?

A very pragmatic one. A repeating item is created by invoking the repeat command for an already created item. Later, it may be edited with the edit command. I want to avoid (more) code duplication  in `ui_day_item_repeat()` and `update_rept()`.

> Is adding a repetition, then editing it the only way to make use of this feature (without importing events)?

The import command does not (yet) support the extensions. But you can edit the `apts` file if you know the file format :-).

> On a related note, I found the prompts for editing repetitions highly confusing and wouldn't know how to use them without reading parts of the code or commit messages.

Or reading RFC5545.

I am very well aware of the problems. The outcome is the result of a process. RFC5545 uses a very general, but powerful notation to describe all sorts of recurring events (section 3.3.10). This has served as a specification for the implementation in commit https://github.com/lfos/calcurse/pull/258/commits/087e192b8d4c213e09f388786dfd37a21eacb586. But a user cannot and should not be expected to know RFC5545.

> I hope we can come up with something that is more intuitive. I'd also be happy if we could keep the process pretty short for "simple" recurrence rules and only show more prompts if the user tells us he wants something more sophisticated. We already ask for the repetition type when adding/editing a repetition. Maybe add one or multiple options for extended repetitions?

Here is a challenge for you: design a user interface that makes the recurrence rule concepts of RFC5545 available in calcurse - in the two-line status bar with one-line input. I couldn't come up with anything. Handwaving is not enough.

For a long time I couldn't see a way other than import of iCalender recurring events. During development I edited the `apts` file directly. The test file was created that way. I don't know the UI of web-based calendar software and whether that could provide inspiration?

I ended with what you see: a direct interface to the three lists that constitute the heart of the extensions. If you understand section 3.3.10 of RFC5545, it is not too difficult to fill them in correctly. Otherwise I would rely on documentation of a suitable number of typical use cases.

lfos:
> A very pragmatic one. A repeating item is created by invoking the repeat command for an already created item. Later, it may be edited with the edit command. I want to avoid (more) code duplication in `ui_day_item_repeat()` and `update_rept()`.

But implementation details should not affect the UI. I guess that means we need to refactor some more code.

> Here is a challenge for you: design a user interface that makes the recurrence rule concepts of RFC5545 available in calcurse - in the two-line status bar with one-line input. I couldn't come up with anything. Handwaving is not enough.

Here's a simple solution: when converting a regular item to a recurrent one, the first prompt asks for a repetition type. In addition to the existing options (daily, weekly, monthly, yearly), we add another option (extended or rfc5545). When using one of the first four options, the interface is exactly like it used to be, without the option to use the extended format. With the new fifth option, we can either show more prompts or just allow the user to enter a repetition description as used in the `apts` file and document that very well.

lhca:
Nothing new. Whether the extended recurrences are available from the repeat command or only from the edit command (when applied to a repeating item) is a minor detail.

lfos:
How is it just a minor detail, considering the end-user experience? We always consistently provide the same options when adding or editing an event. Having to create, then edit an item to get access to a specific feature is very unintuitive...

Any opinion on making the extended format available via an extra type?

lhca:
> Any opinion on making the extended format available via an extra type?

Yes. The extensions are (surprise!) extensions to the basic types, not an extra type. You have to establish type, frequency and until-day before going on.

lfos:
Well, of course you can still ask for the type again afterwards. Alternatively, let the user choose between simple or extended format in a menu before creating the recurrence. I am open for other suggestions, too, but recurrent items should remain usable without having to understand RFC5545.

lhca:
> Alternatively, let the user choose between simple or extended format in a menu before creating the recurrence. I am open for other suggestions, too, but recurrent items should remain usable without having to understand RFC5545.

Proposal.

- The "repeat" command is extended with an opening dialogue that lets the user choose between a simple or an advanced repetition. "Simple" then proceeds as before with selection of type, frequency and until day. "Advanced" is the same, but in addition continues with weekdays, months and monthdays.

- The "advanced" options are documented by a suitable number of typical use cases.

Where should the use case documentation be put? Calcurse has a built-in help system. It seems not to be maintained. Do users use it, I wonder? Other possibilities?

lfos:
Sounds good to me.

Yes, we need to work on improving documentation. Currently, there's the man page, the manual and the built-in help system. Might make sense to get rid of at least one of them.

lhca:
> Sounds good to me.

I will upload a commit shortly. Earlier you wrote

> I'd also be happy if we could keep the process pretty short for "simple" recurrence rules and only show more prompts if the user tells us he wants something more sophisticated.

The dichotomy simple/advanced is in the mind of the user, not a property of the appointment/event. Therefore the user must be asked. This is done as agreed for the create command. But when you edit an existing repeating item, there are prompts for the extensions. In my opinion, a further "simple or advanced?"-dialogue is clumsy and should be avoided. Instead, you type 'Return' when asked for weekdays, months and monthdays.

As for documentation of the extended recurrence rules:

> Currently, there's the man page, the manual and the built-in help system. Might make sense to get rid of at least one of them.

I would vote for dropping the manual. It's not being maintained and while that is so, may do more harm than good.

There is a further possibility: providing help in the status bar (as done in several of the submenus of the 'Edit Item' command). Because it is context dependent, it seems well suited for the recurrence extensions. I am working on it.

lfos:
I agree with everything you said. Thanks for working on this!

lhca:
I'm still working on improved help, but the recurrence rules of RFC 5545 (iCalendar) _are_ complicated. I had a new idea that may make them easier to use:

A new (interactive) command, **next**, which goes to the next occurrence when the selected item in the appointments panel is a repeating one. After creating a new repeating event or appointment, you can use it to check that you got what you intended, simply by invoking it repeatedly and go through the occurrences one by one.

Because it is specialized and we are short of letters for command names, I suggest making it a subcommand of the generic command (default `:`). Then you go through the occurrences by typing `: n RETURN` repeatedly.

lfos:
Great suggestion!

lhca:
The implementation of the **next** command required two small, but important changes not directly related to the recurrence rule extensions. Both are backwards compatible and have been pushed to the PR branch as separate patches.

lhca:
Two more patches uploaded. The final ones.

I have updated the `: help repeat` text with all the interesting(?) details. They should be accessible to an interested user somewhere. I hope they are readable.

The `Repetition` subcommand of `Edit Item` has built-in, context dependent information for the three lists 'Weekdays', 'Months' and 'Monthdays', available if you type '?' as the only input. It is on purpose kept very terse and I have tried to cram everything in, but the two-line status bar does not leave much room for verbose language anyway.

lfos:
No worries, this patch series is not forgotten. I'd like to focus on getting another feature release out before we can focus on this, though. Thanks for the follow-up patches!

lfos:
I tried to merge this but unfortunately, there seem to be conflicts with the iCal patch. I hope it's not too bad. Will look into it soon.

lfos:
I was able to resolve the conflicts with `master` and `pu`. The patch series is currently living on a separate branch. I will do some more testing and reviewing. Expect it to land in `pu` soon.

lfos:
I played around with advanced recurrences a little bit and still had a very hard time understanding how they work and what's possible. I intentionally did not read the RFC, documentation nor the commit message of the main commit, since I wanted to see how useful and easy-to-use the feature is when used intuitively.

After that, I had a look at the documentation which was really helpful. Even though I still have not worked out all the nuances, things make more sense now, great job there!

Maybe it's in the nature of things that this feature won't be usable out of the box, but I still wonder whether we can make it more accessible. However, I believe that everything in this direction would now be considered an improvement on top of this work, so it's not a reason to block the patch series. The user still has the choice to only use simple recurrences and won't ever be confronted with the details.

Even though it's only been a few commits since 4.6.0, I think we have enough material to justify another feature release. I will close the PR but I'd like these patches to stay in `pu` for a while and freeze feature additions for the 4.7.0 branch with the exception of small improvements. That being said, new feature additions are still welcome and will be maintained in a separate branch for the time being.

Thanks again, Lars, for all the amazing work!

PR #264, opened 2020-02-15
Git ignore more

hoijui:
this helps to hide all files auto-generated during the configure and build process (on my system; Debian 10)


lfos:
Merged into pu. Thanks!

PR #266, opened 2020-02-16
Escape contents of SUMMARY field when exporting

kechpaja:
Short-term fix to allow calendar event summaries to contain commas,
semicolons, and colons; refer to RFC 5545
(https://tools.ietf.org/html/rfc5545#page-12) for more details.

Note that this is an imperfect fix. We should also be removing double
quotes that occur in event summaries, but the question of what to
replace them with requires input from more people than me.


lfos:
Why don't we simply escape the problematic characters, as the commit message suggests, instead?

kechpaja:
Going to open a new pull request doing that, but will close this one.

PR #267, opened 2020-02-19
Escape necessary characters in SUMMARY on export

kechpaja:
Escape semicolons, commas, and backslashes when exporting a calendar
item to iCal format.

This should do the necessary escaping in the proper fashion, unlike the previous one.

### DIFF
--- a/src/ical.c
+++ b/src/ical.c
@@ -73,7 +73,7 @@ static const char *ical_recur_type[RECUR_TYPES] =
     { "", "DAILY", "WEEKLY", "MONTHLY", "YEARLY" };
 
 /* Escape characters in field before printing */
-static void escape_value_and_print(FILE * stream, char * field, char * msg)
+static void ical_format_line(FILE * stream, char * field, char * msg)
 {
 	char * p;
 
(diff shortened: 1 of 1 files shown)


lhca:
This looks OK to me.

A small suggestion. On import of an iCalendar file, escaped characters are handled by a function called `ical_unformat_line()`. How about `ical_format_line()` instead of `escape_value_and_print()`? That would also stick with the `ical_*` names of other functions.

@lfos With this patch calcurse produces RFC 5545-conforming ical-files. In the test framework there are a couple of non-conforming input files with commas (ical-002.ical and ical-006.ical). Shouldn't  the tests be corrected? Also, I will raise an issue with `ical_unformat_line()`. It does too much and too little.

kechpaja:
That's a fair point, and an easy change to make. How does it look now? 

I can also change the test files, although that will need to happen a bit later on due to time constraints.

lhca:
> @lfos With this patch calcurse produces RFC 5545-conforming ical-files.

Well, not quite. An ical-file exported by calcurse is non-conforming in two other ways:

- content lines are not delimited by a CRLF line break sequence
- long content lines are not folded

Calcurse import _does_ recognise CRLF as well as folded lines. Should the test input files be amended?

lfos:
Merged into pu, thanks!

PR #270, opened 2020-03-09
Omit duplicate entries when importing ical/ics data: Update llist.c

cecst:
When ical/ics files are imported, omit duplicate entries. Useful for stand-alone calendars on Android and desktops so each can synchronize by exporting/importing their entire calendars to each other repeatedly. The patch prints the details on each entry that is omitted. This behavior could be suppressed with a command-line option.

### DIFF
--- a/src/llist.c
+++ b/src/llist.c
@@ -150,12 +150,29 @@ void llist_add(llist_t * l, void *data)
 }
 
 /*
+ *  return 1 if events are duplicates, 0 otherwise
+ */
+static int event_is_duplicate(struct apoint *a, struct apoint *b)
+{
+        if ((a->start == b->start)          &&
+            (a->dur   == b->dur)            &&
+            ((! a->mesg && ! b->mesg) || 
+             ( a->mesg && b->mesg && (! strcmp(a->mesg, b->mesg))))  &&
+            ((! a->note && ! b->note) || 
+             ( a->note && b->note && (! strcmp(a->note, b->note))))) {
+           return 1;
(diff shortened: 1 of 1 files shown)


lfos:
Thanks for the suggestion! I like the idea. Some questions/comments:

1. Does you current patch work with all types of events? Appointments, (all-day) events, recurrent events and todo items?
2. I'd prefer to not make this the default behavior. Enabling it with a command-line argument sounds like a good idea.
3. There are also some style issues with your contribution. Please try to stick to the coding style that is used everywhere else in our code base.

dennisleexyz:
If anyone is interested in a crude, temporary workaround, I'm currently using an `awk` command in [my pre-load hook](https://github.com/dennisleexyz/dotfiles/blob/master/.config/calcurse/hooks/pre-load) to dedupe on runtime.

lfos:
> If anyone is interested in a crude, temporary workaround, I'm currently using an `awk` command in [my pre-load hook](https://github.com/dennisleexyz/dotfiles/blob/master/.config/calcurse/hooks/pre-load) to dedupe on runtime.

That works. A proper implementation would be nice anyways since this seems to be a common use case.

Strahinja:
Any updates on this? For anyone interested in making a proper implementation of this feature, how would the proposed command line argument be named? Something like `--omit-duplicates`?

lfos:
Closing as this has been inactive for >2 years. Would be glad if somebody would pick this up and complete the work!

Issue #271, opened 2020-03-11
Import of an icalender file


lhca:
Recently @kechpaja (#267) pointed out that an icalendar file exported by calcurse may not conform to RFC 5545 (the proposed iCalendar standard).

Equivalently, calcurse imports non-conforming ical files, possibly making non-standard changes. Examples are

- unescaped commas and semicolons are accepted
- escaped characters other than `\ ; , N` and `n` are accepted
- the character sequence "\n" in SUMMARY is replaced by a space, although RFC 5545 requires a one-line summary without line breaks
- the character sequence "\t" in SUMMARY or DESCRIPTION is replaced by a horisontal tab

Issue #265, opened 2020-02-16
Exported iCalender events not properly escaped


kechpaja:
The iCalendar specifications (https://tools.ietf.org/html/rfc5545#page-11) require that commas, semicolons, colons, and a few other characters be escaped in the values of most fields of an iCalendar event. However, when I export an event from Calcurse that includes commas in the text, they are unescaped, e.g. in the 'SUMMARY' field of the following:

>BEGIN:VCALENDAR
>PRODID:-//calcurse//NONSGML v4.5.1-24-gada0//EN
>VERSION:2.0
>BEGIN:VEVENT
>DTSTART;VALUE=DATE:20200612
>SUMMARY:Absolute, final, no-way-around-asking-for-more-time *graded* thesis deadline
>UID:89455665661fc33f4ab85ab7b1df8ce07594301f
>END:VEVENT
>END:VCALENDAR

The unescaped commas are then misinterpreted by my CalDAV server, which truncates the summary of the above event to "SUMMARY:Absolute".

kechpaja:
#266 provides a very simple patch that wraps the contents of the SUMMARY field in quotes when an event is exported, allowing events containing commas, semicolons, and colons to meet the specifications. However, it does nothing to prevent double quotes (forbidden by RFC 5545) from making it into the exported iCalendar item — I decided that adding code to remove or replace them would be overstepping my bounds as a random contributor sharing a fix that I personally had to make, since there are a couple of options for how to deal with them and the decision about which one is correct should be made by the project maintainers.

lhca:
> The iCalendar specifications (https://tools.ietf.org/html/rfc5545#page-11) require that commas, semicolons, colons, and a few other characters be escaped in the values of most fields of an iCalendar event.

This is a very broad, inaccurate statement, and the reference is misleading. We must look at the property in question: SUMMARY. The value of a SUMMARY property is of type TEXT which is defined here: https://tools.ietf.org/html/rfc5545#section-3.3.11.

lhca:
> #266 provides a very simple patch that wraps the contents of the SUMMARY field in quotes when an event is exported, allowing events containing commas, semicolons, and colons to meet the specifications.

I don't see how quoting should help. In a value of type TEXT, the double quote (DQUOTE) is an allowed character.

> However, it does nothing to prevent double quotes (forbidden by RFC 5545) 

? Please  provide a reference.

kechpaja:
[...]

PR #273, opened 2020-03-22
iCalendar import

lhca:
Issue #265 led to a minor review of import of iCalendar files, and I raised issue #271.

This PR comprises five patches. One adresses #271, while the others
- improve and fix import logging
- fix display of interactively imported todos
- add three new tests

Note that two of the tests will fail until PR #267 has been accepted.


test/ical-010.sh
```diff
@@ -0,0 +1,34 @@
...
+  cp "$DATA_DIR/apts-export" .calcurse/apts || exit 1
+  cp "$DATA_DIR/todo-export" .calcurse/todo || exit 1
+  "$CALCURSE" -D "$PWD/.calcurse" --export=ical | sed '/^PRODID/d'
+  rm -rf .calcurse || exit 1
+elif [ "$1" = 'expected' ]; then
+  cat <<EOD
+BEGIN:VCALENDAR
+VERSION:2.0
+BEGIN:VEVENT
+DTSTART:20200226
```
lfos: Seems like this was generated/tested with a version of calcurse that does not include commit 0114289 (ical: use the VALUE parameter in DTSTART, 2018-05-09). I fixed this locally.

Issue #281, opened 2020-04-24
calcurse-caldav: recur-events can't be pushed to a server (error with the formatting of the DTSTART property)


Jerem-K:
Using calcurse-caldav from calcurse 4.6.0, I can't push recur-events to the server. 
I've used the SyncFilter and calcurse-caldav works fine with event, apt or recur-apt.

For example, I got this message for a recur-event:

Pushing new object 6ccf86b33a4a75cd2b13211124c57a30f328caf3 to the server.
error: The server at *** replied with HTTP status code 415
error: (Unsupported Media Type) while trying to access https://***.***
error: .***/remote.php/dav/calendars/***/***/6ccf86b33a4a75cd2b132111
error: 24c57a30f328caf3.ics.

I used the debug option and I got this:
<   <s:message>Validation error in iCalendar: The supplied value (20200424) is not a correct DATE-TIME</s:message>

I investigated further and I saw that the DTSTART property is not correctly written for recur-events:
DTSTART:20200424 instead of DTSTART;VALUE=DATE:20200424 (this is what happens with "normal" events)

lhca:
You are absolutely right. A similar problem for events where reported a long time ago (#81 and #94) and solved by #97. But the latter omitted recurring events.

A fix is upcoming.

Jerem-K:
Thanks @lhca 

I created a pull request that fixes this issue: #282

PR #282, opened 2020-04-25
Fix formatting of DTSTART for recurrent events

Jerem-K:
Solve issue #281

Thanks @lhca for the answer and the tip!

More testing is needed.


lhca:
Looks OK to me. Unfortunately it cannot be tested by importing into calcurse which at the moment simply ignores the value parameter (`;VALUE=DATE`). I have a patch in the pipeline that will make calcurse less forgiving.

lfos:
Looks good to me. Merged into `pu`. Thanks for the quick fix!

PR #294, opened 2020-06-01
Import and recurrence rules update

lhca:
The PR consists of six commits pertaining to import of iCalendar files and/or recurrence rules.

Time zone identifiers in import files are now recognized. Date-times are converted to local date-times of the system running calcurse. An import note with the time zone identifier is added to the appointment. Start time conversion for an event with recurrence rules may result in an appointment that cannot be imported because the start time no longer matches the rules.

Calcurse now supports import of RFC 5545 recurrence rule parts BYDAY, BYMONTH and BYMONTHDAY. Some import templates are available in contrib/import and may serve as starting points for use of these advanced rules.

The interactive UI of the repeat/edit-repetition command has been extended so that the until date may be set with a repeat count.

The configuration option "general.systemdialogs" has been removed. The "Welcome to Calcurse" window is no longer displayed, while the import progress report is always displayed; it may still be suppressed with the command line option -q.


lfos:
Merged into pu, thanks!

lhca:
Note: unmerged commits.

lfos:
Which commits are unmerged?

lhca:
On Sun, Jun 14, 2020 at 06:20:29AM -0700, Lukas Fleischer wrote:
> 
>    Which commits are unmerged?

Well, that's what Github says. And commit 12514ea is the latest on lfos/pu.

lfos:
I see. Forgot to push to the GitHub remote, should be fixed now. Thanks for the pointer.

PR #296, opened 2020-06-09
Fix typos

tbm:
Signed-off-by: Martin Michlmayr <tbm@cyrius.com>


tbm:
The a/an issues can be found with [anorack](https://github.com/jwilk/anorack).

Issue #289, opened 2020-05-18
[Feature Request] Show notes when viewing a task


JPTIZ:
Started using calcurse today and it is actually pretty nice and convenient :) But speaking of convenience...

An opened task window is almost the size of the whole application, but it only shows the title (which is expected to be oneliner - at least in my understanding), while for notes there is a separate command to show notes (but it doesn't show the task's title).

I think it would make more sense to actually show the notes in that big task window, so I can quickly check the notes without having to type `>`. The `>` could be left to use when notes become too big to fit the window.

Adam-Antios:
Yes! Exactly that! It is peculiar for this to not be the normal behaviour!

rosalogia:
I tried implementing this and ran into some minor problems, and furthermore am unsure as to whether I am doing this responsibly since I don't really write C.

You can see which parts are my implementation from the little green `+`s indicating additions on my end:
![todonoteimpl](https://user-images.githubusercontent.com/66163965/94519236-41a85480-01f8-11eb-8e48-acd91ed05e9c.png)

Here's how it looks in use:
![notepic](https://user-images.githubusercontent.com/66163965/94519454-816f3c00-01f8-11eb-94ff-1a75aa2427ec.png)

However, I'm finding that closing the TODO view (with `v`) and opening it again in place (with `v` again) causes this little rendering bug (see the little `^?` at the bottom?):
![notebug](https://user-images.githubusercontent.com/66163965/94519536-b11e4400-01f8-11eb-905e-a8d8f77da8db.png)

I have no clue what's causing it, so any pointers on where to start looking for the issue would be appreciated.

JPTIZ:
Wow, nice job, @rosalogia! :) 

At first glance it __looks__ like some null-terminated string issue, although you do seem to take care of it. Can you double-check it? If it is the case, I think it would happen in that `msg` variable.

Also, what's the branch you're working on so I can try something here?

rosalogia:
Thank you!

I will happily double check, but you can take a look on [my fork](https://github.com/rosalogia/calcurse/commit/aa811d9dd5495031c95af98822f160767bc18386).

rosalogia:
[...]

PR #322, opened 2020-10-01
Display Notes in TODO View and Event/Appt. View

rosalogia:
This PR implements the feature requested in #289 and expands this functionality to events and appointments as well. When viewing TODO items or Events/Appts., if one exists, the associated note is displayed in the pop-up window and is truncated after reaching a specified character limit.

EDIT: I would like to be clear about the fact that I'm very inexperienced with C, and would appreciate any effort to review or further test the code from anyone willing to help :smile:

### DIFF
--- a/src/day.c
+++ b/src/day.c
@@ -582,8 +582,30 @@ void day_write_stdout(time_t date, const char *fmt_apt, const char *fmt_rapt,
 void day_popup_item(struct day_item *day)
 {
 	if (day->type == EVNT || day->type == RECUR_EVNT) {
-		item_in_popup(NULL, NULL, day_item_get_mesg(day),
-			      _("Event:"));
+		if (day_item_get_note(day)) {
+			size_t note_size = 3500;
+			char note[note_size];
+
+			char *notepath;
+			FILE *fp;
+			asprintf(&notepath, "%s%s", path_notes, day_item_get_note(day));
+
+			fp = fopen(notepath, "r");
+			mem_free(notepath);
+
+			note_read_contents(note, note_size, fp);
+			fclose(fp);
+
+			char *msg;
+
+			asprintf(&msg, "%s\n\nNOTES:\n%s", day_item_get_mesg(day), note);
+
+			item_in_popup(NULL, NULL, msg, _("Event:"));
+
+			mem_free(msg);
+		} else {
+			item_in_popup(NULL, NULL, day_item_get_mesg(day), _("Event:"));
+		}
 	} else if (day->type == APPT || day->type == RECUR_APPT) {
 		char a_st[100], a_end[100];
 
@@ -593,8 +615,30 @@ void day_popup_item(struct day_item *day)
 		apt_tmp.start = day->start;
 		apt_tmp.dur = day_item_get_duration(day);
 		apoint_sec2str(&apt_tmp, ui_day_sel_date(), a_st, a_end);
-		item_in_popup(a_st, a_end, day_item_get_mesg(day),
-			      _("Appointment:"));
+
+		if (day_item_get_note(day)) {
+			size_t note_size = 3500;
+			char note[note_size];
+
+			char *notepath;
+			FILE *fp;
+			asprintf(&notepath, "%s%s", path_notes, day_item_get_note(day));
+
+			fp = fopen(notepath, "r");
+			mem_free(notepath);
+
+			note_read_contents(note, note_size, fp);
+			fclose(fp);
+
+			char *msg;
+
+			asprintf(&msg, "%s\n\nNOTES:\n%s", day_item_get_mesg(day), note);
+			item_in_popup(a_st, a_end, msg, _("Appointment:"));
+
+			mem_free(msg);
+		} else {
+			item_in_popup(a_st, a_end, day_item_get_mesg(day), _("Appointment:"));
+		}
 	} else {
 		EXIT(_("unknown item type"));
 		/* NOTREACHED */
--- a/src/ui-todo.c
+++ b/src/ui-todo.c
@@ -314,7 +314,30 @@ void ui_todo_popup_item(void)
 	if (!item)
 		return;
 
-	item_in_popup(NULL, NULL, item->mesg, _("TODO:"));
+	if (item->note) {
+		/* Assign a sane default note size that will cleanly
+		 * truncate long notes */
+		size_t note_size = 3500;
+		char note[note_size];
+
+		char *notepath;
+		FILE *fp;
+		asprintf(&notepath, "%s%s", path_notes, item->note);
+
+		fp = fopen(notepath, "r");
+		mem_free(notepath);
+		note_read_contents(note, note_size, fp);
+		fclose(fp);
+
+		char *msg;
+
+		asprintf(&msg, "%s\n\nNOTES:\n%s", item->mesg, note);
+
+		item_in_popup(NULL, NULL, msg, _("TODO:"));
+		mem_free(msg);
+	} else {
+		item_in_popup(NULL, NULL, item->mesg, _("TODO:"));
+	}
 }
 
 void ui_todo_flag(void)
--- a/src/calcurse.h
+++ b/src/calcurse.h
@@ -1019,6 +1019,7 @@ void edit_note(char **, const char *);
 void view_note(const char *, const char *);
 void erase_note(char **);
 void note_read(char *, FILE *);
+void note_read_contents(char *, size_t, FILE *);
 void note_gc(void);
 
 /* notify.c */
--- a/src/note.c
+++ b/src/note.c
@@ -155,6 +155,17 @@ void note_read(char *buffer, FILE * fp)
 	buffer[MAX_NOTESIZ] = '\0';
 }
 
+/* Read the contents of a note file */
+void note_read_contents(char *buffer, size_t buffer_len, FILE * fp)
+{
+	size_t read_count = fread(buffer, 1, buffer_len, fp);
+	if (read_count != buffer_len)
+		buffer[read_count] = '\0';
+	else
+		memcpy(&buffer[buffer_len - 4], "...\0", 4);
+}
+
+
 static void
 note_gc_extract_key(struct note_gc_hash *data, const char **key, int *len)
 {


lfos:
Thanks a lot Ambika! Your approach looks good overall but I do have a few comments regarding the implementation:
* Is there a reason we're limiting the note output to 3500 characters? What happens if we read the whole file?
* There seem to be whitespace/indentation issues; particularly in the asprintf() line. Please also make sure each commit is properly formatted on its own (as opposed to introducing inconsistent indentation and fixing it in a followup commit).
* Please try to avoid strcpy() and any buffer length computations. For example, instead of using snprintf() you can use asprintf() which automatically allocates a buffer that is large enough to hold the result. No careful computations needed.

rosalogia:
Hi @lfos! Sorry for the late reply.

> Is there a reason we're limiting the note output to 3500 characters? What happens if we read the whole file?

Not particularly! But the pop-up window will automatically truncate text after a certain point (without a trailing `...`) so I felt that reading in the whole file might have been wasteful and introduce unnecessary complexity.

I found that it's quite difficult to predict at what point text will be automatically truncated, since it is not dependent at all on number of lines or length of lines. Instead, text automatically wraps, and one very very long line or many very very short lines will cause auto-truncation. I am not sure that there is a way to calculate how a hard maximum that actually corresponds to how text is rendered in the pop-up, so I went with a number that didn't truncate too early for large text files.

Of course, if we were to read in the whole file we could just let the pop-up window handle truncation, and although we would be missing the trailing `...`, they are arguably unnecessary. However, I am not sure whether or not this would be the most resource-efficient way of going about this.

> There seem to be whitespace/indentation issues; particularly in the asprintf() line.

I noticed this initially, and I'm not sure whether or not you noticed that in the second commit on the PR these issues were fixed (as far as I'm aware).

> Please also make sure each commit is properly formatted on its own (as opposed to introducing inconsistent indentation and fixing it in a followup commit).

So would you rather I rebase and squash these two commits into one so that a commit with improper formatting isn't introduced to the branch history? If so, that's something I can do!

> Please try to avoid strcpy() and any buffer length computations. For example, instead of using snprintf() you can use asprintf() which automatically allocates a buffer that is large enough to hold the result. No careful computations needed.

This is absolutely doable. Thanks!

lfos:
> Of course, if we were to read in the whole file we could just let the pop-up window handle truncation, and although we would be missing the trailing `...`, they are arguably unnecessary. However, I am not sure whether or not this would be the most resource-efficient way of going about this.

Thanks for the explanation. There are probably more sophisticated ways of achieving the same thing in a resource efficient yet more generic/consistent way (e.g. using the product of the numbers of rows and the number of columns of the screen as a upper bound for the contents read from the file, then adding the ellipsis in the rendering code) but the hardcoded limit seems like a good start.

> > There seem to be whitespace/indentation issues; particularly in the asprintf() line.
> 
> I noticed this initially, and I'm not sure whether or not you noticed that in the second commit on the PR these issues were fixed (as far as I'm aware).

I did but it's always preferrable to keep commits atomic and clean (both for readers of the history and for certain Git commands operating on individual commits).

> > Please also make sure each commit is properly formatted on its own (as opposed to introducing inconsistent indentation and fixing it in a followup commit).
> 
> So would you rather I rebase and squash these two commits into one so that a commit with improper formatting isn't introduced to the branch history? If so, that's something I can do!

It should be fairly straightforward to uncommit your second commit (e.g. using `git reset HEAD^`), stage only the whitespace changes (e.g. using `git add -p`), then merge the whitespace changes into the first commit (e.g. using `git commit --amend`), then recreate the second commit only with the non-whitespace changes (e.g. using `git commit -a`).

> > Please try to avoid strcpy() and any buffer length computations. For example, instead of using snprintf() you can use asprintf() which automatically allocates a buffer that is large enough to hold the result. No careful computations needed.
> 
> This is absolutely doable. Thanks!

Thank you!

rosalogia:
Apologies for the repeated force-pushes :grimacing: I'm not sure there's a better way to amend history that's been pushed already but I definitely could've minimised the quantity. Each time I thought I was done I noticed something else I had forgotten :sob:

That being said, I'm a little unsure as to how I might do away with the use of `strcpy` in [L165](https://github.com/lfos/calcurse/pull/322/commits/d2c64fe57f7765ebcd26fc6cd01890eedcfcb548#diff-e128ceb99ac8eb7d8ad1ec858b833fa1R165) of `note.c`. What I'm trying to achieve here is replace the final few elements of the buffer with `"...\0"` if the file exceeds maximum length. Could this be done with `asprintf` or a related function?

lfos:
No worries. In that specific case, you could either use asprintf() with a length specifier to copy 3497 characters, then append the `...`, or use your approach; I'd suggest using memcpy() instead of strcpy() with what you're doing, just to make it more explicit that the command also adds the terminating NUL-character, though.

JPTIZ:
Finally got my chance to review this :) I can see that there's been a lot of discussion here already. So, here: I tried to compile the code with your PR, but I'm stuck at this:

```
day.c: In function ‘day_popup_item’:
day.c:594:4: error: implicit declaration of function ‘note_read_contents’ [-Werror=implicit-function-declaration]
  594 |    note_read_contents(note, note_size, fp);
      |    ^~~~~~~~~~~~~~~~~~
```

(I have `-Werror=implicit-function-declaration` flag set globally for reasons of my work, which triggers this as an error, but it is originally a warning)

I would suggest to declare the function in `calcurse.h` (there's a list of function declarations for `ui-day.c` around line 1145).

rosalogia:
I believe I have handled both of these concerns! Please let me know if there's anything I missed :smile:

src/ui-todo.c
```diff
@@ -314,7 +314,28 @@ void ui_todo_popup_item(void)
...
-	item_in_popup(NULL, NULL, item->mesg, _("TODO:"));
+	if (item->note) {
+		/* Assign a sane default note size that will cleanly
+		 * truncate long notes */
+		size_t note_size = 3500;
+		char note[note_size];
+
+		char *notepath;
+		FILE *fp;
+		asprintf(&notepath, "%s%s", path_notes, item->note);
```
lfos, line 325: Memory leak, please run mem_free(notepath) when notepath is no longer needed.

src/ui-todo.c
```diff
@@ -314,7 +314,28 @@ void ui_todo_popup_item(void)
...
+
+		fp = fopen(notepath, "r");
+		note_read_contents(note, note_size, fp);
+		fclose(fp);
+
+		char *msg;
+
+		asprintf(&msg, "%s\n\nNOTES:\n%s", item->mesg, note);
```
lfos, line 333: Memory leak, please run mem_free(msg).

src/ui-todo.c
```diff
@@ -314,7 +314,28 @@ void ui_todo_popup_item(void)
...
+
+		item_in_popup(NULL, NULL, msg, _("TODO:"));
+	}
+	else
```
lfos, line 337: Style: Please put else on the same line as the closing braces; use braces for an else-block if the if-block has braces (even if the body consists of just a single line). This line should be "} else {".

src/day.c
```diff
@@ -582,8 +582,28 @@ void day_write_stdout(time_t date, const char *fmt_apt, const char *fmt_rapt,
...
 	if (day->type == EVNT || day->type == RECUR_EVNT) {
-		item_in_popup(NULL, NULL, day_item_get_mesg(day),
-			      _("Event:"));
+		if (day_item_get_note(day)) {
+			size_t note_size = 3500;
+			char note[note_size];
+
+			char *notepath;
+			FILE *fp;
+			asprintf(&notepath, "%s%s", path_notes, day_item_get_note(day));
```
lfos, line 591: Memory leak. Please fix (see comments in previous patch for details).

src/day.c
```diff
@@ -582,8 +582,28 @@ void day_write_stdout(time_t date, const char *fmt_apt, const char *fmt_rapt,
...
+
+			fp = fopen(notepath, "r");
+			note_read_contents(note, note_size, fp);
+			fclose(fp);
+
+			char *msg;
+
+			asprintf(&msg, "%s\n\nNOTES:\n%s", day_item_get_mesg(day), note);
```
lfos, line 599: Memory leak. Please fix.

src/day.c
```diff
@@ -593,8 +613,28 @@ void day_popup_item(struct day_item *day)
...
-		item_in_popup(a_st, a_end, day_item_get_mesg(day),
-			      _("Appointment:"));
+
+		if (day_item_get_note(day)) {
+			size_t note_size = 3500;
+			char note[note_size];
+
+			char *notepath;
+			FILE *fp;
+			asprintf(&notepath, "%s%s", path_notes, day_item_get_note(day));
```
lfos, line 623: Memory leak. Please fix.

src/day.c
```diff
@@ -593,8 +613,28 @@ void day_popup_item(struct day_item *day)
...
+
+			fp = fopen(notepath, "r");
+			note_read_contents(note, note_size, fp);
+			fclose(fp);
+
+			char *msg;
+
+			asprintf(&msg, "%s\n\nNOTES:\n%s", day_item_get_mesg(day), note);
```
lfos, line 631: Memory leak. Please fix.

src/day.c
```diff
@@ -582,8 +582,28 @@ void day_write_stdout(time_t date, const char *fmt_apt, const char *fmt_rapt,
...
+
+			item_in_popup(NULL, NULL, msg,
+					_("Event:"));
```
lfos, line 602: Style: No line break, append to previous line.

src/day.c
```diff
@@ -582,8 +582,28 @@ void day_write_stdout(time_t date, const char *fmt_apt, const char *fmt_rapt,
...
+		}
+		else
```
lfos, line 604: Style: Move else to line above and add braces (see previous patch for details).

src/day.c
```diff
@@ -593,8 +613,28 @@ void day_popup_item(struct day_item *day)
...
+			note_read_contents(note, note_size, fp);
+			fclose(fp);
+
+			char *msg;
+
+			asprintf(&msg, "%s\n\nNOTES:\n%s", day_item_get_mesg(day), note);
+			item_in_popup(a_st, a_end, msg,
+					_("Appointment:"));
+		}
+		else
```
lfos, line 635: Style: Move else to line above and add braces (see previous patch for details).

src/day.c
```diff
@@ -593,8 +613,28 @@ void day_popup_item(struct day_item *day)
...
+
+			fp = fopen(notepath, "r");
+			note_read_contents(note, note_size, fp);
+			fclose(fp);
+
+			char *msg;
+
+			asprintf(&msg, "%s\n\nNOTES:\n%s", day_item_get_mesg(day), note);
+			item_in_popup(a_st, a_end, msg,
+					_("Appointment:"));
```
lfos, line 633: Style: No line break, append to previous line.

lfos (commented):
Thanks for the revision! I only have a few minor comments; see inline.

rosalogia:
Thanks :smile: I hadn't consciously registered that once I started using `asprintf` I was allocating dynamically. Hopefully I've addressed all your concerns, but please do let me know if there's anything I missed!

src/ui-todo.c
```diff
@@ -314,7 +314,30 @@ void ui_todo_popup_item(void)
...
+		 * truncate long notes */
+		size_t note_size = 3500;
+		char note[note_size];
+
+		char *notepath;
+		FILE *fp;
+		asprintf(&notepath, "%s%s", path_notes, item->note);
+
+		fp = fopen(notepath, "r");
+                mem_free(notepath);
```
lfos, line 328: Whitespace issue. Please fix.

src/day.c
```diff
@@ -582,8 +582,30 @@ void day_write_stdout(time_t date, const char *fmt_apt, const char *fmt_rapt,
...
+			fclose(fp);
+
+			char *msg;
+
+			asprintf(&msg, "%s\n\nNOTES:\n%s", day_item_get_mesg(day), note);
+
+			item_in_popup(NULL, NULL, msg, _("Event:"));
+
+			mem_free(msg);
+		} else  {
```
lfos, line 606: Style nit: Duplicate space.

lfos (commented):
I think we're getting really close. I found two more small whitespace issues. Looks good otherwise, thanks!

rosalogia:
This should do it :) Thanks for all your help and of course do let me know if there's anything that could be fixed or improved upon!

JPTIZ:
Yay, nice, seems to work perfectly here :)

image

lfos:
Merged into pu, thanks!

(By the way, I noticed that your last name is different in the Author header of the two patches. Is that intended?)

rosalogia:
It is not intended :sweat_smile: Some of the information in my local git configuration is/was outdated and I must have updated some of it while this PR was being worked on. It's not much of a bother to me, but if it's easy to fix maybe that's worth doing!

lhca:
Sorry for this belated comment.

The text `NOTES:` should be translatable and not hard coded. Except for the word "TODO", which is special, capitalized forms are not used elsewhere. I would prefer "Note:" as default.

lfos:
Agreed, good catch Lars. I amended the patches in `pu`.

lfos:
Ambika, I also updated your author information in the first patch (assuming the information in the second patch was correct). If you would like to change it, please let me know.

Issue #328, opened 2020-11-11
[FEATURE REQUEST] Completion status for whole-day appointments


machinedgod:
Hello,
the way I usually manage my daily tasks is not via task list, but via daily appointments that have no set time:
![2020-11-11-093926_968x544_scrot](https://user-images.githubusercontent.com/5933618/98825142-28ebab00-2402-11eb-8f45-0e46013dba39.png)

I manually type in the `[ ] <task name>` as description, then edit the task to add an `x` as a checkmark.
If I don't manage to complete the task(s), I copy/paste them to another day and then delete them.

I only use tasklist for long-term tasks that I, at some point in time, have to remember to do. At that point, I schedule them into my day and remove them from the task list.

I am not sure if this is even acceptable feature, depending on the calcurse philosophy, but it'd be super useful for me (and I hope anyone else) if these three features would get in:
- daily tasks (appointments without specific time, with a finished/unfinished state)
- promotion of tasks from task list, to the daily task (selected day in calendar)
- moving the task `+-1day`, `+-1w`, `+-1m` and `+-1y` (I couldn't figure out how to do this without copy pasting)

I'm fairly capable C coder, and I need this feature enough that I'd be willing to put it in myself, but I do need some guidance as to what is the best way to do this. Specifically:
a) which datatype to look at as an example
b) which datatype to extend with new functionality, or where to introduce a new one
c) what would be acceptable key shortcuts for task promotion and task movement

IF this feature is unacceptable to calcurse, I am still hoping to get some help since in that case, I'd be maintaining  a fork (which, if really necessary, could be private as well).

Sincerely

machinedgod:
Just a note - started working on it on my fork, have promotion part done, will add the rest shortly. I used 'P' key when TODO window is in focus, it seems not to be used anywhere.

Code is so easy to navigate through and change, I'm very impressed!

lhca:
Thank you for the interesting idea. Just to avoid any mistake, your "tasklist" is what in `calcurse` is called the to-do list, and your "daily task" is an event (as opposed to an appointment). Right?

You propose

* daily tasks (appointments without specific time, with a finished/unfinished state)
* promotion of tasks from task list, to the daily task (selected day in calendar)
* moving the task `+-1day`, `+-1w`, `+-1m` and `+-1y` (I couldn't figure out how to do this without copy pasting)

[...]

PR #330, opened 2020-11-12
New feature: promote TODO items to APP items and toggleable APP events

machinedgod:
#328 
* Add new key handler 'P' that promotes TODO item into an event on a currently
  selected day. New event is allocated, TODO item mesg pointer is copied into
  it, then TODO item is disposed of (obviously, a bug)
  It also automatically turns it into a 'toggleable' event, which just means
  it prepends `[ ] ` to the message
* If "flag" (!) key is used on events in APP, it'll do one of the three things
  to the `mesg`:
    - if it starts with `[ ] `, replace it with `[x] `.
    - if it starts with `[x] `, replace it with `[ ] `.
    - if it starts with anything else, reallocate a new buffer of size
      strlen(mesg) + 4, copy original message into it and prepend a `[ ] `.

### DIFF
--- a/src/calcurse.c
+++ b/src/calcurse.c
@@ -211,7 +211,13 @@ static inline void key_repeat_item(void)
 static inline void key_flag_item(void)
 {
 	if (wins_slctd() == APP && !event_dummy(ui_day_get_sel())) {
-		ui_day_flag();
+		struct day_item *item = ui_day_get_sel();
+		if(item->type == EVNT) {
+			ui_day_item_toggle_check();
+		} else {
+			ui_day_flag();
+		}
+		
 		do_storage(0);
 		wins_update(FLAG_APP);
 	} else if (wins_slctd() == TOD) {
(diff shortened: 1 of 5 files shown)


machinedgod:
As mentioned in the issue, please feel free to remark anything, from implementation details to code style.

I tried to review everything myself, but its _possible_ I left space indents somewhere. I keep `expandtabs` option ON by default, and didn't realize until the end that its all tabs. Tried to fix it, but some might've slipped under. Sorry on that.

pkillnine:
This is great, thanks for making this. One small thing that isn't a very big deal is that some of the created appointments have extra redundant characters at the end, like `V^B`.

Also, it would be nice if you could demote a todo-appointment back to a todo-item.

I like the simplicity of making a todo item be just a normal appointment but with `[ ]` at the beginning.

machinedgod:
> This is great, thanks for making this. One small thing that isn't a very big deal is that some of the created appointments have extra redundant characters at the end, like `V^B`.
> 
> Also, it would be nice if you could demote a todo-appointment back to a todo-item.
> 
> I like the simplicity of making a todo appointment be just a normal appointment but with `[ ]` at the beginning.

Thanks for the comment :-)
Yeah, I'm aware of sometimes garbage being copied with the task name - I think the len calculation is a bit off, will have to take a look into that.

Apropos demoting the event back to tasks, we were having a bit larger discussion on it in the Issue related to this, the project owners had some really good ideas - so we'll probably go down that route. I'll try and do it that way when I get some time.

lfos:
Closing as there hasn't been any activity for >1 year. Would be more than happy if somebody would pick this up, polish, and (re-)submit!

PR #335, opened 2020-12-03
calcurse-caldav: allow non-ascii characters in username and password

Grimler91:
Hi, 

I was trying to `calcurse-caldav` with a radicale server where password contained non-ascii characters, and got an error:
```
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-2: ordinal not in range(128)
```
With this patch syncing works as expected.


lfos:
Merged into pu, thanks!

Issue #257, opened 2020-01-07
[FEATURE REQUEST] Support for 12 Hour Time


rajansaini691:
Using version 4.5.1-8-g74f4-dirty. 
I am trying to change calcurse to use the 12 hour clock instead of the 24 hour clock. Is this feature supported? Thanks!

lhca:
When run interactively calcurse uses the hour (time of day) in three places, two for output, one for input.

In the notify bar at the bottom of the screen, the calcurse clock outputs the time (and date). The format can be configured in 'strftime()' fashion to your hearts content. Use the 'n Notify' submenu of the 'C Config' menu.

In the appointments panel, calcurse outputs the time for each appointment. The format is hardcoded as a 24-hour clock (hh:mm). It would not be too difficult to make it configurable.

In several situations calcurse asks for time-of-day input (e.g. when a new appointment is entered). Only a 24-hour format is accepted (hh:mm or hhmm).

lfos:
How about adding a `outputime` format option to the `general` section? It could then be used in both interative and non-interactive mode (and can be overwritten in non-interactive mode by using extended format specifiers).

JakeRoggenbuck:
Yeah, I think adding a setting to make it 12 hour time would be very helpful

PR #339, opened 2020-12-24
Allow configuration of appointment time format

mercurialmoon:
Added the option to configure the format in which appointment time is
displayed. The setting is called "format.appointmenttime" under the
general settings menu. Setting defaults to previous behavior, which was
"%H:%M".

This addresses issue #257


lfos:
Merged into `pu`, thanks!

PR #343, opened 2021-01-18
New feature: Cycle backwards trough windows.

Aurora2500:
Added new key handler `KEY_BTAB` (Shift-Tab) that cycles backwards trough the windows `(TODO -> APP -> CAL)`.


src/keys.c
```diff
@@ -69,6 +69,7 @@ static struct keydef_s keydef[NBKEYS] = {
 	{ "generic-copy", "c", gettext_noop("Copy") },
 	{ "generic-paste", "p ^V", gettext_noop("Paste") },
 	{ "generic-change-view", "TAB", gettext_noop("Chg Win") },
+	{ "generic-change-view-p", "KEY_BTAB", gettext_noop("Prev Win") },
```
lfos, line 72: Is there a particular reason this is named `generic-change-view-p` (vs. `generic-change-view-prev`)? Is there a length limitation somewhere?

lfos (commented):
Looks good (one minor comment inline). I'll merge this for the next feature release; thanks for your contribution!

src/keys.c
Aurora2500: There is. Under the config menu, going to the keys configuration, having it be `generic-change-view-prev` seems to overflow over the `=`, looking like
```
generic-change-view-pr=vKEY_BTAB
```

lfos: `generic-prev-view`?

Aurora2500: Sounds good to me, might also be a good idea in that case to change all the `generic-change-view`to `generic-next-view` for consistency.

lfos: Thanks! The problem is the label is used in the config file too. Changing an existing label will cause an error when loading an existing key file unless a workaround is added.

Aurora2500: I see, seems like it's getting out of scope of what the original pull request was about. I'll revert that for now and make an maybe issue on that so that it could be looked at in due time.

I wouldn't really know what else would need to be changed as simply grepping the whole source dir doesn't give any more hits for `generic-change-view`.

src/calcurse.c
```diff
@@ -80,6 +80,13 @@ static inline void key_generic_change_view(void)
 	wins_update(FLAG_ALL);
 }
 
+static inline void key_generic_change_view_prev(void)
```
lfos, line 83: Nit: Could you please also rename the function (to key_generic_prev_view())...

src/calcurse.c
```diff
@@ -851,6 +858,7 @@ int main(int argc, char **argv)
 		wtimeout(win[KEY].p, -1);
 		switch (key) {
 		HANDLE_KEY(KEY_GENERIC_CHANGE_VIEW, key_generic_change_view);
+		HANDLE_KEY(KEY_GENERIC_CHANGE_VIEW_PREV, key_generic_change_view_prev);
```
lfos, line 861: ... and the enum value to KEY_GENERIC_PREV_VIEW? Thanks!

lfos:
Squashed all commits into a single commit and merged into `pu`, thanks!

Issue #346, opened 2021-01-23
[Feature request] Tagging appointments and to-do items


pkillnine:
This is like a lighter version of the multiple calenders feature request - if you can add tags to appointments and to-do items, you could show only appointments/todo's that have particular tags, so you could tag items with 'work' or 'personal'. And also have multiple tags, so you can kind of easily sub-categorise them.

This would be really nice for storing all kinds of appointments too, you could have a calender of potential events to go to without mixing them up with your actual commitments.

lfos:
I like the proposal and would be happy to receive patches.

I also think that, with this implemented, there's probably not too much work left to add support for multiple calendars (but I might be wrong on this).

goblin:
I was just about to post a similar request.

I was thinking the tags could have a few configurable items:
* flag text: a character (or maybe two, max three) that would appear next to the appointment or todo item, similar to `!` or `X.`  appearing when you flag with `!`
* flag color: the color of that flag text
* event color: would color the whole event in the chosen colors.
* key binding

The tags could be toggled via a configurable key binding, similar to vim's marks, i.e. `mw` could toggle the "work" tag, `mp` the "personal" tag etc. Unfortunately, `m` is currently used for "+1 month", and `t` for "+1 day", so that would need to be adjusted somehow.

Finally, the tags could be exported to iCal using the `CATEGORIES` property.

goblin:
As above, I've implemented a simple tagging support for now.

I've also created a small converter utility that can print events colorized according to tags: https://github.com/goblin/ical2svg

PR #352, opened 2021-02-06
Add very crude proof-of-concept tagging support

goblin:
I've implemented very crude and basic tagging to address https://github.com/lfos/calcurse/issues/346. I'm not sure if it's pull-worthy but I hope it can be useful to someone until we get full-blown tagging support. It may also be useful as a guide to someone willing to do more work on this.

Only non-recurring, non-full-day appointments (i.e. not the ones considered "events" in the code) can be tagged by using the `'` key followed by a single character. Pressing the same key combination again removes the tag. Up to 8 tags can be stored (this can be changed by modifying `TAGS_MAX` in calcurse.h:354). Tags should be saved and loaded from the `apts` file, and will also be exported to iCal as CATEGORIES. Importing from iCal isn't implemented yet. Colouring isn't implemented either.

This is useful enough for me, so I'm not planning to do much more work on this.


goblin:
(tags are just single characters for now, so you can use things like "w" for "work" and "p" for "personal")

lfos:
Thanks, this is a good start. For a full implementation, it'd be great if we could have support for all types of items (and, if possible, also allow more tags and/or more verbose tags); I'm inclined to not include this in mainline as-is since it can be confusing if the feature only works on a quite limited set of items. Hopefully somebody will pick this up and add the missing parts. I'll leave this open for the time being.

lfos:
No activity for more than one year. Closing; I'd still be more than happy to merge if somebody would pick this up and polish!

Issue #356, opened 2021-03-14
calcurse-caldav deletes its own items


Andonome:
# Summary

When running calcurse-caldav to synchronize with efoundation's nextcloud, the first run pushes the new appointment up, the second run pulls it down and deletes the local version, and the third run deletes the remote version.

# Dots

My /caldav/config is [here](https://gitlab.com/andonome/dots/-/blob/new/skel/config/calcurse/caldav/config)

All hooks have been disabled for testing.

# Steps to replicate

1. Set up dot files with your nextcloud
2. Push to Nextcloud (maybe this is just efoundation?)
2. Make efoundation calendar entry
3. calcurse-caldav --init=keep-remote
4. Make appointment in calcurse
5. calcurse-caldav
6. Repeat step 5 three times.
7. Notice the appointment has been deleted from everywhere.

The attached file has the output of `calcurse-caldav --debug`.

[calcurse-log.txt](https://github.com/lfos/calcurse/files/6137851/calcurse-log.txt)

next-mad-hatter:
If I'm not mistaken this is another manifestation of https://github.com/lfos/calcurse/issues/337 -- the calendar url in the attached log contains the "@" character, hence this (incorrect) behaviour.

lfos:
Thanks for the report; this is something we should definitely address. I replied in #337.

lfos:
I merged @next-mad-hatter's PR into `pu`.

@Andonome, could you please confirm that the issue no longer occurs when using `calcurse-caldav` from `pu`?

Andonome:
@lfos 
Tested on calcurse 4.7.1 -- text-based organizer.

Everything working on Arch Linux and Arch Linux Arm.

lfos:
Thanks!

Issue #337, opened 2020-12-13
Fastmail breaks calcurse-caldav


next-mad-hatter:
Problem is reproducible with current master (e3dd92e -- contents of calcurse-caldav is identical to 4.7.0 release version afaics).

When the href path contains url-special characters, calcurse-caldav cannot establish correct mapping between local and remote (possibly only those which were created locally) calendar items; this results in items being erroneously deleted or created.

To reproduce, create a calendar entry locally, then sync with a CalDAV server with a url-special character in the href path -- in particular, Fastmail puts `/dav/calendars/user@domain/` in there.  Upon next synchronization the local item will be removed, and upon yet next one the remote item as well.

As far as I can see, this is due to the path currently being escaped in return value of `push_object`, but not being escaped anywhere else, especially in `get_etags`.  I can see correct behaviour with `urllib.parse.quote` removed, but have not tested for possible regressions.

lfos:
Interesting, thanks for the report!

Note that we didn't quote the `href` originally, the current behavior was introduced in commit e943b06 (URL encode href before saving to sync.db, 2017-09-06). @watersalesman might be able to provide more context.

Looking at the implementation, I wonder whether we should quote `href` earlier in push_object() instead, i.e., quote before using it anywhere instead of only quote for the sync DB entry.

@next-mad-hatter Could you try moving the urllib.parse.quote() to where `href` is originally defined (first line in push_object()) and don't quote in the return value? (You may have to reinitialize the sync DB and/or run at least twice after the change.)

next-mad-hatter:
Merged https://github.com/lfos/calcurse/pull/358 solves this for me

PR #358, opened 2021-04-04
When pushing local objects, always take href as it is returned by server

next-mad-hatter:
Apparently, while some caldav servers will return href values as is in their response (as seen in https://github.com/lfos/calcurse/issues/337 and https://github.com/lfos/calcurse/issues/356), some might return them url-escaped (which if I am not mistaken was the reason for https://github.com/lfos/calcurse/commit/e943b0605ff7000280bcf190f35f9097a6ae5f67).

Assuming either behaviour when pushing local objects will lead to discrepancy with events dictionary retrieved from the server and thus bugs, hence we always want to use whatever form of href the server returns.

### DIFF
--- a/contrib/caldav/calcurse-caldav.py
+++ b/contrib/caldav/calcurse-caldav.py
@@ -9,7 +9,6 @@ import re
 import subprocess
 import sys
 import textwrap
-import urllib.parse
 import xml.etree.ElementTree as etree
 
 import httplib2
@@ -425,13 +424,17 @@ def push_object(conn, objhash):
     headerdict = dict(headers)
     if 'etag' in headerdict:
         etag = headerdict['etag']
-    while not etag:
+
+    returned_href = None
+    while not etag or not returned_href:
         etagdict = get_etags(conn, [href])
         if etagdict:
-            etag = next(iter(etagdict.values()))
-    etag = etag.strip('"')
+            returned_href, returned_etag = next(iter(etagdict.items()))
+            if not etag:
+                etag = returned_etag
 
-    return (urllib.parse.quote(href), etag)
+    etag = etag.strip('"')
+    return (returned_href, etag)
 
 
 def push_objects(objhashes, conn, syncdb, etagdict):


contrib/caldav/calcurse-caldav.py
```diff
@@ -421,17 +420,14 @@ def push_object(conn, objhash):
...
-    headerdict = dict(headers)
-    if 'etag' in headerdict:
-        etag = headerdict['etag']
-    while not etag:
-        etagdict = get_etags(conn, [href])
-        if etagdict:
-            etag = next(iter(etagdict.values()))
+    etagdict = get_etags(conn, [href])
+    if not etagdict:
+        return None
```
lfos, line 425: A few questions:

1. Does this introduce a possible race condition when there's another update after pushing the new item but before retrieving the ETag? Should we only use the response to update `href` but not the ETag (unless it is missing)?
2. This also seems to remove the loop. Can you add a comment explaining why it is removed and how this impacts the behavior of the script?

lfos (commented):
Thanks for the PR!

Is this change tested?

Other than that, I only have two more questions/comments inline and one nit: Could you please change the commit message subject line to something short that includes `calcurse-caldav` (e.g. `calcurse-caldav: always take href from server`) and include some of the information you added in your comment here to the commit message body? Thanks again!

contrib/caldav/calcurse-caldav.py
next-mad-hatter: Thank you for the review,

yes, this does simplify the logic too much, I'll push a more conservative edit which should again account for those two points.

next-mad-hatter:
@lfos Rebased to account for the two points raised as well as commit message.

I've done some ad hoc testing against fastmail servers so far, but won't be able to do much more at the moment.  If someone has time to test these changes against other providers or add end-to-end tests, that would be fantastic of course, but unfortunately outside of my resource pool right now.

For what it's worth this could hopefully serve as a starting point in solving #337 and #356.

lfos:
Merged into `pu` with a few small changes, thanks a lot!

PR #370, opened 2021-05-22
Make Home/End keys work as expected in text input mode

cryptolmath:
Users expect these keys to move the cursor rather than input nonsensical strings.


lfos:
Merged into pu, thank you!

PR #371, opened 2021-05-22
Make PgUp/PgDown skip back/ahead one week

cryptolmath:
In the current default configuration these keys do nothing. Presumably, most users would consider skipping one week reasonable behaviour.


lfos:
Thank you for your contribution! As there aren't a lot of unmapped keys left, I would (slightly) prefer not adding additional keys to actions that already have defaults. I am also unsure whether having PgUp/PgDown jump by a week (vs. a month/year) makes sense in general. If anybody feels strongly that those keys should be mapped by default, please reopen, happy to revisit.

PR #385, opened 2021-07-09
Fix hook docs

DhruvaSambrani:


lfos:
Merged into pu. Thank you!

PR #387, opened 2021-08-09
fix typo in caldav post-sync hook git commit message

josephmturner:
Calcurse is great! I appreciate your hard work.

I tried submitting this patch with git send-email (sent to bugs at calcurse dot org), but I think it may not have been received.

Gratefully,

Joseph


lfos:
Merged into pu, thank you!

Issue #393, opened 2021-09-13
Calcurse-caldav creates unnecessary directories in `XDG_DATA_HOME` and `XDG_CONFIG_HOME`


josephmturner:
First off, thank you! This program works extremely well for me.

I'm running commit 2b76604 (tip of `origin/pu`)

My calcurse data directory structure looks like

```
$HOME/.config/calcurse/
├── family
│   ├── apts
│   ├── caldav/
│   ├── conf
│   ├── daemon.log
│   ├── hooks/
│   ├── notes/
│   └── todo
└── personal
    ├── apts
    ├── caldav/
    ├── conf
    ├── daemon.log
    ├── hooks/
    ├── notes/
    └── todo
```

My calcurse config directory has a similar structure inside `$HOME/.config/calcurse/{family,personal}`.

I run `calcurse-caldav` with the appropriate `--config`, `--syncdb`, and `--datadir` arguments, and everything works smoothly, save for one small issue.

Each time it runs, `calcurse-caldav` creates `caldav` directories inside `$HOME/.local/share/calcurse/` and `$HOME/.config/calcurse`. This messes with my dmenu script for choosing which calendar to open.

If we were to add a `--configdir` option to `calcurse-caldav`, we could set `caldav_config` (line 605) to that instead of `$XDG_CONFIG_HOME`. But if we tried to use the argument passed to `--datadir` to set `caldav_data` (line 606), that would be a breaking change.

I'm not sure what a clean solution would look like. Maybe some environment variables (`CALCURSE_CALDAV_DATA` and `CALCURSE_CALDAV_CONFIG`) would allow us to set `caldav_config` and `caldav_data` before `check_dir` ever gets called.

Curious to hear your thoughts. Thank you very much!

Joseph

PR #394, opened 2021-09-17
Add env variables CALCURSE_CALDAV_DATA and CALCURSE_CALDAV_CONFIG

josephmturner:
Allows users to set arbitrary locations for caldav config and data,
without the calcurse-caldav script creating unused `caldav` directories
inside $XDG_DATA_HOME/calcurse and $XDG_CONFIG_HOME/calcurse.

Addresses #393


lfos:
Thanks for your contribution, Joseph! If I'm not mistaken, all paths can already be configured using command-line arguments (--config, --hookdir, ...) is there anything that's not covered by them?

You're correct that using them will still lead to standard directories being generated to check_dir() calls. I think we should move the check_dir() calls to a later point in the script, after command line arguments have been parsed. Does that make sense?

PR #399, opened 2021-10-16
sha1: Fix checksum calculation on big-endian systems

### LINKED ISSUE #397: ical-003.sh fails on s390x
**Version information.**

calcurse-4.7.1

**Bug description.** 

While upgrading the [Alpine Linux](https://alpinelinux.org) Edge calcurse package to the latest version (4.7.1) I noticed that the calcurse test suite fails with 4.7.1 on s390x, on all other architectures it passes successfully. Furthermore, it seems to have passed successfully on s390x with 4.6.0 in the past.

**Reproduce.**

s390x is a somewhat obscure architecture. The easiest way to reproduce this is probably setting up a qemu-s390x Alpine Edge chroot with binfmt_misc or using a qemu-system-s390x Alpine VM. In either case, one just needs to run `make && make check` to reproduce this issue.

**Expected Behavior.**

All tests should pass.

**Actual Behavior.**

All tests expect `ical-003.sh` pass:

```
$ cat test/test-suite.log
=========================================
   calcurse 4.7.1: test/test-suite.log
=========================================

# TOTAL: 77
# PASS:  76
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: ical-003.sh
=================

Running ./ical-003.sh... FAIL
FAIL ical-003.sh (exit status: 1)
```

More information:

```
calcurse-4.7.1/test$ ./ical-003.sh actual > /tmp/actual
calcurse-4.7.1/test$ ./ical-003.sh expected > /tmp/expected
calcurse-4.7.1/test$ diff -u /tmp/actual /tmp/expected
--- /tmp/actual
+++ /tmp/expected
@@ -210,12 +210,12 @@
  - ..:.. -> 00:00
 	First weekend in May
 First weekend in May is a two-day event!
-Repeating event, three years.
---
-Import: multi-day event changed to one-day event
-First weekend in May is a two-day event!
 Repeating appointment.
 The first weekend in May is a two-day event.
 Non-repeating event.
+--
+Import: multi-day event changed to one-day event
+First weekend in May is a two-day event!
+Repeating event, three years.
 --
 Import: multi-day event changed to one-day event
```

If you need further information to debug this, let me know. I will try to provide it.

nmeum:
Instead of using a custom macro which must be set manually to determine
host byte order (WORDS_BIGENDIAN) this commit uses the __BYTE_ORDER__
CPP macro which is commonly available on many systems [\[1\]][1]. If the macro
is not defined, little endian byte order is silently assumed which is
what the previous version of this code did anyhow. Alternatively, it
would also be possible to emit a #warning if __BYTE_ORDER__ is not set
or rewrite the code in a way that it does not depend on the native host
byte order in the first place [\[2\]][2].

This fixes the calcurse test suite on s390x.

Fixes #397

[1]: https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html
[2]: https://commandcenter.blogspot.com/2012/04/byte-order-fallacy.html

---

CC: @lhca

### DIFF
--- a/src/sha1.c
+++ b/src/sha1.c
@@ -47,7 +47,11 @@
 
 #define rol(val, n) (((val) << (n)) | ((val) >> (32 - (n))))
 
-#ifdef WORDS_BIGENDIAN
+/* We check for host system byte order using the __BYTE_ORDER__
+ * CPP macro which is available on many platforms but not mandated by
+ * the POSIX or C standard. As such, if __BYTE_ORDER__ is not set, we
+ * silently assume little endian byte order. */
+#if __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
 #define blk0(i) block->l[i]
 #else
 #define blk0(i) (block->l[i] = (rol (block->l[i], 24) & \


lfos:
Thank you, good catch!

Since we're already using Autoconf, could we add the AC_C_BIGENDIAN macro to configure.ac instead, which should auto-detect endianness and set WORDS_BIGENDIAN accordingly [1]? (Without having taken a closer look, my guess it that it's going to use an implementation very similar to yours but may handle additional corner cases.)

[1] https://www.gnu.org/software/autoconf/manual/autoconf-2.60/html_node/C-Compiler.html

lhca:
The title and solution is misleading (although it solves the problem). This is _not_ a SHA-1 implementation problem, but a Calcurse _build_ problem.

As already mentioned, configure.ac should be modified by adding `AC_C_BIGENDIAN`, and sha1.c should be left alone.

jordan31bit:
Interesting issue, sadly I have no clue about autoconf tools or writing macros. Though, I am currently reading up on it just to better understand the issue and solution.

PR #405, opened 2021-11-30
updated copyright year from 2020 to 2021

Ghibranalj:
its almost 2022 and the copyright is still 2020.
Ill set a reminder to change it to 2022 in january :D


lfos:
Thanks Ghibran! I just pushed `01ad848` to the pu branch, which updates copyright ranges in all files to 2022.

PR #435, opened 2022-06-22
Display empty event description

### LINKED ISSUE #422: Empty all-day events are hard to notice / deleting the description has a different from trying to create an empty event
*Before filing a new bug report, make sure you use the search function to check whether the bug has already been reported.*

**Version information.**

4.7.1 but I checked the commits since then, I don't think this is fixed yet

**Bug description.**
If you create an all-day event event (no time) with a description, then edit this description to delete it, you end up with an invisible event. I only figured it out because I saw that there was an empty line somewhere.

**Expected Behavior.**
Either:
* the empty event should be somehow visible
* it shouldn't be possible to fully delete the description
* deleting the description should delete the event (e.g. if you try to create a new event with no date and no description, nothing is done)

**Screenshots.** 
In this screenshot, there's an invisible event above the two visible events.
2022-04-19-222618_1219x134_scrot

jonakeys:
This PR addresses issue #422 When an event description is empty, it now shows '(empty description)'.

Fixes #422 

Signed-off-by: Jonathan van der Steege <jonathan@jonakeys.nl>

### DIFF
--- a/src/calcurse.h
+++ b/src/calcurse.h
@@ -307,7 +307,7 @@ struct conf {
 };
 
 #define EMPTY_DAY_DEFAULT "--"
-#define EMPTY_EVENT_DESC_DEFAULT "(empty description)"
+#define EMPTY_EVENT_DESC_DEFAULT _("(empty description)")
 
 /* Daemon-related configuration. */
 struct dmon_conf {
--- a/src/calcurse.h
+++ b/src/calcurse.h
@@ -307,6 +307,7 @@ struct conf {
 };
 
 #define EMPTY_DAY_DEFAULT "--"
+#define EMPTY_EVENT_DESC_DEFAULT "(empty description)"
 
 /* Daemon-related configuration. */
 struct dmon_conf {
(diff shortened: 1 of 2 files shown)


src/calcurse.h
```diff
@@ -307,6 +307,7 @@ struct conf {
 };
 
 #define EMPTY_DAY_DEFAULT "--"
+#define EMPTY_EVENT_DESC_DEFAULT "(empty description)"
```
lfos, line 310: Thanks for your contribution and sorry for the slow response. Could you please make this string localizable (i.e., _("(empty description)") instead of "(empty description)").

src/day.c
```diff
@@ -198,11 +198,17 @@ static void day_add_item(int type, time_t start, time_t order, union aptev_ptr i
...
+	{
 	case APPT:
 		return day->item.apt->mesg;
 	case EVNT:
-		return day->item.ev->mesg;
+		message = day->item.ev->mesg;
+		if (*message != '\0')
+			return message;
+		else
+			return EMPTY_EVENT_DESC_DEFAULT;
```
lfos, line 211: Nit: Prefer to return early -

    if (message[0] == '\0')
        return _("(empty description)");
    return message;

jonakeys:
Added changes as requested by @lfos 
I have no problem with your response time. It's your project and we'll follow your tempo!

lfos:
Merged, and added a follow-up commit to extend this to all item types (including recurrent items and TODO items). Thanks, Jonathan!

PR #436, opened 2022-07-01
calcurse-caldav: Support PasswordCommand option

rlue:
This commit adds a new `Auth/PasswordCommand` option
to support security best practices re: handling secrets
in CLI program configuration.

Prior to this commit, the two available options
for specifying a password were:

  1. via the `Auth/Password` config parameter, or
  2. via a `$CALCURSE_CALDAV_PASSWORD` environment variable.

The former is unsafe for obvious reasons;
the latter is unsafe because as long as the script is running,
its environment can be accessed via

    $ cat /proc/<pid>/environ

and is thus visible to anyone with access to the system.

This commit preserves preexisting behavior (for backward compatibility)
but removes all mention of option 2 from the README.
Since the README example for option 2 used a password command anyway,
there is little reason to continue its use,
and this commit recommends it be deprecated.

### DIFF
--- a/contrib/caldav/calcurse-caldav.py
+++ b/contrib/caldav/calcurse-caldav.py
@@ -6,6 +6,7 @@ import configparser
 import os
 import pathlib
 import re
+import shlex
 import subprocess
 import sys
 import textwrap
@@ -30,6 +31,7 @@ class Config:
         self._map = {
             'Auth': {
                 'Password': None,
+                'PasswordCommand': None,
                 'Username': None,
             },
             'CustomHeaders': {},
@@ -657,9 +659,6 @@ verbose = args.verbose
 debug = args.debug
 debug_raw = args.debug_raw
 
-# Read environment variables
-password = os.getenv('CALCURSE_CALDAV_PASSWORD')
-
 # Read configuration.
 config = Config(configfn)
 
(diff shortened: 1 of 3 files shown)


supermarin:
@lfos any chance you can :eye: this one? Needed this option today and stumbled upon this one

contrib/caldav/calcurse-caldav.py
```diff
@@ -674,7 +672,14 @@ def run_hook(name):
...
 verbose = verbose or config.get('General', 'Verbose')
 
-password = password or config.get('Auth', 'Password')
+if os.getenv('CALCURSE_CALDAV_PASSWORD'):
+    # This approach is deprecated, but preserved for backwards compatibility
+    password = os.getenv('CALCURSE_CALDAV_PASSWORD')
+elif config.get('Auth', 'Password'):
+    password = config.get('Auth', 'Password')
+elif config.get('Auth', 'PasswordCommand'):
+    password = subprocess.run(config.get('Auth', 'PasswordCommand').split(' ', 1), capture_output=True).stdout.decode('UTF-8')
```
lfos, line 681: This only work correctly in corner cases. Could you please change this to tokenize the command with shlex.split()? Please also ensure the password variable gets initialized if none of the three cases apply (i.e., no password is provided).

Everything else LGTM, thanks for your contribution and sorry for the delayed review!

rlue: Thanks so much for your feedback; the latest changes have been applied in a force-push. :pray:

lfos:
Merged, thank you!

PR #465, opened 2023-08-05
Feature request 425 implemented

youhajjioui:


lfos:
Thanks for the PR! I don't think we should dictate the file format for notes; many editors can be already configured to switch to Markdown formatting based on file path. If we'd still like to change the file extension, I believe it should be configurable. Patches to add a configurable extensions are welcome!

PR #468, opened 2023-08-12
Fix segmentation fault when exporting ical file

bwildenhain:
If a note file gets accidently deleted this leads to ical_export_note trying to execute fclose(NULL) which leads to a segmentation fault on current Debian GNU/Linux using libc6 (according to fclose's manpage, the behaviour is undefined in this case).

The fix tests whether fopen returned a non-NULL-pointer before trying to close it.


lfos:
Merged, thank you!

PR #485, opened 2024-03-15
Update flag.txt flag documentation

NotRexButCaesar:
This describes the actual behavior.  If we want to implement the other behavior, it makes sense to change the documentation once the behavior changes.


lfos:
Merged into the `pu` branch, thank you!

PR #502, opened 2025-02-26
Update README for asciidoc

NWuensche:


lfos:
Merged into `pu` with amended commit message. Thanks!

PR #503, opened 2025-03-20
Fix notification command zombie processes

### LINKED ISSUE #390: No SIGCHLD handler, notify-send becomes zombie after being executed as notification.command
I searched the source code and I didn't see SIGCHLD being handled anywhere. This causes forked processes turning into zombies. Consider adding a SIGCHLD handler. It could be something like the following:
```
void
sigchld(int unused)
{
	while (waitpid(-1, NULL, WNOHANG) > 0);
}
```

stephan49:
This PR fixes #390 by waiting for state changes in child processes from the notification command, which prevents zombie processes from accumulating.

### DIFF
--- a/src/sigs.c
+++ b/src/sigs.c
@@ -85,6 +85,9 @@ static void generic_hdlr(int sig)
 		want_reload = 1;
 		ungetch(KEY_RESIZE);
 		break;
+	case SIGCHLD:
+		wait(NULL);
+		break;
 	}
 }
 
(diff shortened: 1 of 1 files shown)


lfos:
Thanks for the PR!

That said,  it looks like this may be a regression introduced in commit f743eab (Remove SIGCHLD signal handler, 2020-07-14). Instead of trying to come up with a novel fix, we should take a careful look at that commit and rationale in the commit message, then decide whether that commit should be reverted.

PR #504, opened 2025-04-23
calcurse-caldav: use flock to prevent concurrent instances

Flying-Toast:
The current locking implementation has two issues:
1. There is a race condition between checking for lockfile existance and creating it
2. A process holding the lock can die without removing the file

This commit fixes both problems with a new locking implementation. To obtain the lock, a process obtains an exclusive flock on the lockfile, creating the file if it doesn't exist. The kernel takes care of removing the flock once the locked descriptor is closed (i.e. when the process exits).

### DIFF
--- a/contrib/caldav/calcurse-caldav.py
+++ b/contrib/caldav/calcurse-caldav.py
@@ -3,6 +3,7 @@
 import argparse
 import base64
 import configparser
+import fcntl
 import os
 import pathlib
 import re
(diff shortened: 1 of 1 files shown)


lfos:
Merged into the `pu` branch, thanks for the contribution!

themarcman314:
I am encountering issues with the lock consistently not being deleted. For now I'm just ignoring the check for locks, but I was wondering if this PR addresses that or maybe I have a configuration problem.

Thanks!

Flying-Toast:
That’s the intended behavior with this change. The concept of being “locked” changes from “does the file exist?” to “is there currently a process that has an active advisory lock on the file?” (see https://www.man7.org/linux/man-pages/man2/flock.2.html). The purpose of the file here is merely to serve as an anchor point for contesting calcurse instances to call flock() on.

themarcman314:
Ok thanks!
I appreciate you taking the time to answer.

PR #514, opened 2025-08-08
Change iCal default extension from "ical" to "ics"

OneTrueC:
RFC 5545 Section 8.1 [1] specifies that the extension for iCalendar
files is "ics":

> File extension(s):  The file extension of "ics" is to be used to
> designate a file containing (an arbitrary set of) calendaring and
> scheduling information consistent with this MIME content type.

[1] https://www.ietf.org/rfc/rfc5545.txt

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -79,7 +79,7 @@ static FILE *get_export_stream(enum export_type type)
 	const char *wrong_name =
 	    _("The file cannot be accessed, please enter another file name.");
 	const char *press_enter = _("Press [ENTER] to continue.");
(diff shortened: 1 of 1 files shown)


lfos:
Thanks! I think this is a good idea overall. Have you considered backwards-compatibility? What'd happen with workflows with calcurse iCal exports that may depend on the existing extension?

OneTrueC:
(apologies for the month+ delay)

The function where this change is is only used interactively so I can't imagine that there are many workflows depending on it. The only real recourse I could see for this (should it be a problem at all) is to version bump/denote it as a breaking change.

lfos:
Makes sense. Merged into `pu`, thanks!

Issue #172, opened 2019-01-22
caldav digest auth


muelli:
The calendar I am trying to synchronise does only use digest auth, rather than basic auth. I couldn't make the caldav sync script work and from digging into the source, I think I've found the culprit:

https://github.com/lfos/calcurse/blob/870fa1aa327c443979bddfc862a691597b8a2273/contrib/caldav/calcurse-caldav.py#L149

Other calendaring implementations handle digest auth well by way of a standard HTTP library. I'm thinking of Evolution and DavDroid in particular.

PR #516, opened 2025-09-01
Digest authentication support for CalDav

melpomene:
Added support for CalDav digest authentication. 

Tested on a basic Baikal setup with the default authentication method. 

Issue #172


lfos:
Thanks for the PR! Does httplib2 not support digest authentication out of the box? (At least part of the documentation seems to suggest it does.) If it doesn't, is there another simple and widely used library we can use instead of implementing the logic ourselves?

melpomene:
Good feedback, thank you. I'll have a look.

PR #524, opened 2026-01-09
Update apoint.c

factorial52:
Switch to -year, month, day- format.


scowcron:
format strings would need to be changed as well.

PR #526, opened 2026-04-11
Ignore empty CalDAV entry to allow syncing to continue

foolswood:
I don't entirely know why the CalDAV server I was using came back with a response like this, but failing to encode the None as UTF-8 didn't feel like desirable behaviour.

If we just ignore it and continue then the sync works great, so that seemed like the simplest thing to do.

Don't know if this is affecting anyone else, but presumably it could and hopefully the change isn't too scary.

### DIFF
--- a/contrib/caldav/calcurse-caldav.py
+++ b/contrib/caldav/calcurse-caldav.py
@@ -537,6 +537,11 @@ def pull_objects(hrefs_missing, hrefs_modified, conn, syncdb, etagdict):
             die_atnode('Missing calendar data.', node)
         cdata = cdatanode.text
 
+        if cdata is None:
+            if verbose:
+                print("Ignoring empty object {}.".format(etag))
+            continue
+
         if href in hrefs_modified:
             if verbose:
                 print("Replacing object {}.".format(etag))


foolswood:
Realised it would probably be better to show a minimal example response that triggers this so you can see what's actually happening.

```
<?xml version="1.0" encoding="UTF-8"?>
<D:multistatus xmlns:D="DAV:" xmlns:CR="urn:ietf:params:xml:ns:carddav" xmlns:A="http://apple.com/ns/ical/" xmlns:K="Kolab:" xmlns:CS="http://calendarserver.org/ns/" xmlns:C="urn:ietf:params:xml:ns:caldav">
  <D:response>
    <D:href>/calendars/nope@notgivingthat.away/calendar/</D:href>
    <D:propstat>
      <D:prop>
        <D:getetag>1728205163</D:getetag>
      </D:prop>
      <D:status>HTTP/1.1 200 OK</D:status>
    </D:propstat>
    <D:propstat>
      <D:prop>
        <C:calendar-data xmlns:C="urn:ietf:params:xml:ns:caldav"/>
      </D:prop>
      <D:status>HTTP/1.1 404 Not Found</D:status>
    </D:propstat>
  </D:response>
</D:multistatus>
```

